PR #10, opened 2022-10-21
Add clustering method for basins fractions continuation

awage:
This PR is the port of the PR in ChaosTools.jl/#279.


Datseris:
@awage please make sure lines of code don't exceed 92 characters. If you have automatic text wrapping in your editor, I'd recommend to turn it off. The PR doesn't have any tests, can we add something basic, and not heavy on performance, to make sure the code works? We can use e.g., the magnetic pendylum continuation I also have at the recurrences method

awage:
It is still ongoing work. I am testing on the Kuramoto for consistency with the paper but I will implement some basic test units.

src/continuation/continuation_clustering.jl
```diff
@@ -0,0 +1,133 @@
...
+    # Cluster them
+    metric = cc.clust_distance_metric
+    # @show features
+    f = reduce(hcat, features) # Convert to Matrix from Vector{Vector}
+    f = float.(f)
+    features_for_optimal = if cc.max_used_features == 0
+        f
+    else
+        StatsBase.sample(f, minimum(length(features), cc.max_used_features); replace = false)
+    end
```
Datseris, line 119: this wont work. Here `sample` will not give you a Matrix. It will give you a vector. You first need to sample from `features` before you convert them into `f`.

Datseris: Also, please do not do `float.(f)` that's unecessary computations. The feature vectors should be given as floats to begin with. If not, it's the users problem.

Datseris:
This is what you want for the Mmap directory: https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.mktemp-Tuple{AbstractString}

awage:
> This is what you want for the Mmap directory: [https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.mktemp-Tuple{AbstractString}](https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.mktemp-Tuple%7BAbstractString%7D)

Thanks! I am testing the code with the kuramoto network. As soon as I have the results I update the code and finish the PR (at last!)

awage:
@Datseris I am done with this PR, I have the results of the Kuramoto oscillators and they are compatible with the results of the paper for the U(-pi,pi) case: 
kur_continuation_mcbb_pi_pi_9000

Captura de pantalla_2022-10-25_21-28-10

There is a test for the continuation recurence method included in this PR. I though it would be useful but if you don't think it is necessary I can revert this change.

awage:
Thanks for the feedback. I have some problem with the optimal radius. I am still debugging. Also I have some ideas to modify the interface to allows different types of mappers.

src/continuation/continuation_clustering.jl
```diff
@@ -0,0 +1,170 @@
...
+    # Extract the first possible feature to initialize the features container
+    feature = extract_features(mapper, ics; N = 1)
+    features = Vector{typeof(feature[1])}(undef, n*spp)
+    # Collect features
+    for (i, p) in enumerate(prange)
+        set_parameter!(mapper.integ, pidx, p)
+        current_features = extract_features_threaded(mapper, ics; show_progress, N = spp)
+        features[((i - 1)*spp + 1):i*spp] .= current_features
+        next!(progress)
+    end
```
Datseris, line 88: This must become its own function called something like `extract_features_across_parameters`. Furthermore, you should use `extract_features` instead of `extract_features_threaded`. The latter is called if `threaded = true`.

src/continuation/continuation_clustering.jl
```diff
@@ -0,0 +1,170 @@
...
+    return means ./ N
+end
+
+
+
+function basins_fractions_continuation(
+        continuation::NamedTuple, prange, pidx, ics::Function;
+        samples_per_parameter = 100, show_progress = true, par_weight = 1, 
+        ϵ_optimal = 1., mmap_limit = 20000
+    )
```
Datseris, line 72: All the keywords besides `samples_per_parmeter, show_progress` should instead be given s arguments to the struct `ClusteringAcrossParametersContinuation`. Then, use that struct in the `basins_fractions_continuation` method.

Datseris (commented):
About the `ϵ_optimal = 1.`. The field `clusterconfig` of `mapper` has information of how to obtain the optimal ε, right? So why are we allowing this keyword? Shouldn't the user specify how to get ε directly in the mapper config? Then, we don't have to redo stuff here, and we just deduce the ε using the existing mapper code. Additionally, in the docstring we should strongly recommend users to either pick the optimal radius method to be either an explicitly given real number, or to put an upper limit in the amount of features used to get optimal ε.

src/continuation/continuation_clustering.jl
```diff
@@ -0,0 +1,158 @@
...
+            means[i] += f[i]
+        end
+    end
+    return means ./ N
+end
+
+
+
+function basins_fractions_continuation(
+        continuation::NamedTuple, prange, pidx, ics::Function;
```
Datseris, line 70: `NamedTuple` vcan be anything; we have to make a specific type `ClusteringAcrossParametersContinuation`, not just a function that returns a named tuple.

awage:
Closed for PR #25

PR #12, opened 2022-10-25
Make sparse the default mode in AttractorsViaRecurrences

KalelR:


test/mapping/attractor_mapping_tests.jl
```diff
@@ -74,12 +74,12 @@ function test_basins(ds, u0s, grid, expected_fs_raw, featurizer;
         test_basins_fractions(mapper; known = true, err = aerr)
     end
 
-    @testset "Recurrences" begin
-        mapper = AttractorsViaRecurrences(ds, grid; diffeq, show_progress = false, kwargs...)
+    @testset "Recurrences Non-Sparse" begin
+        mapper = AttractorsViaRecurrences(ds, grid; sparse=false, diffeq, show_progress = false, kwargs...)
```
Datseris, line 78: Since the non sparse and sparse version have identical code, and the only thing that changes is an internal array represenrtation, we shouldnt have to test the non sparse version extensively. I propose that you remove the non sparse from these large, detailed tests, and just add a new test elsewere that computes some basins fractions with sparse and non sparse and tests that the result is identical between the two.

KalelR:
I added a test comparing the two versions for Lorenz84. It is currently failing because the versions give slightly different results:
E.g. `Dict(2 => 0.246, -1 => 0.003, 3 => 0.211, 1 => 0.54)` for sparse versus `Dict(2 => 0.222, -1 => 0.003, 3 => 0.219, 1 => 0.556)` for nonsparse.  

Should I relax the restrictions such that the tests pass?

test/mapping/recurrence_tests.jl
```diff
@@ -0,0 +1,37 @@
...
+        ds = Systems.lorenz84(; F, G, a, b)
+        u0s = [
+            1 => [2.0, 1, 0], # periodic
+            2 => [-2.0, 1, 0], # chaotic
+            3 => [0, 1.5, 1.0], # fixed point
+        ]
+        diffeq = (alg = Vern9(), reltol = 1e-9, abstol = 1e-9)
+        M = 200; z = 3; xg = yg = zg = range(-z, z; length = M); grid = (xg, yg, zg)
+        test_compatibility_sparse_nonsparse(ds, grid; diffeq)
+    end
```
Datseris, line 36: Let's use a cheaper system here, the Henon map preferably.

KalelR:
There are still two unrelated errors: one related to `trajectory_continuous` and another caused by an outdated Entropies.jl version used in these tests, which doesnt have Renyi().

test/mapping/recurrence_tests.jl
```diff
@@ -0,0 +1,34 @@
+using Attractors
+using Test
+using OrdinaryDiffEq:Vern9
```
Datseris, line 3:
```suggestion
```

src/basins/fractality_of_basins.jl
```diff
@@ -139,6 +139,7 @@ function basins_fractal_test(basins; ε = 20, Ntotal = 1000)
     return tst_res, Ŝbb
 end
 
+linreg(x, y) = hcat(fill!(similar(x), 1), x) \ y #as suggested in https://github.com/JuliaStats/StatsBase.jl/issues/398#issuecomment-417875619
```
Datseris, line 142:
```suggestion
# as suggested in https://github.com/JuliaStats/StatsBase.jl/issues/398#issuecomment-417875619
linreg(x, y) = hcat(fill!(similar(x), 1), x) \ y
```

Datseris (approved):
Remaining failing tests are fine. I have to take care of them. Or you can, by reading the new Entropies.jl docs. 

Go ahead and commit the two changes I outlined and then you can merge!

Datseris:
@KalelR for future reference: never "merge" requests via merge commit. Always choose "Squash and merge". There is a dropdown arrow in the merge button that allows the squash version. Always do squash.

KalelR:
Ah, ok! Sorry, will do!

PR #13, opened 2022-10-25
(WIP) Denser attractors mapped by Recurrences

KalelR:
Two additions to make the attractors found by Recurrences denser:
1. add `stop_at_dt` in step! 
2. store the attractors every time a cell if revisited, so that the number of points in attractors will be at least `mx_chk_loc_att` 

This is one fix for the problem of the algorithm finding two limit cycles when it should find only one. We found this occurs when the limit cycles stored in memory are "sparse", in the sense that there are large gaps between their points. The change No 2. does not by itself fix the problem. Something happens sometimes when `stop_at_dt=false` that makes the integrator only reach the same set of points belonging to the limit cycle and it then does not fill it densely. It seems from some quick tests that this is not only when `\Delta t` is commensurate with the cycle's period. There may be something with how it is stepping, but I don't understand that.

One point to verify is if these changes slow the code down significantly. And whether it is good to store so many points in the attractors. At the very least we should probably improve the documentation on this keyword, as it seems both George and I were confused about it (though the code itself is quite clearly written).

### DIFF
--- a/src/mapping/attractor_mapping.jl
+++ b/src/mapping/attractor_mapping.jl
@@ -3,7 +3,6 @@
 
 export AttractorMapper,
     AttractorsViaRecurrences,
-    AttractorsViaRecurrencesSparse,
     AttractorsViaProximity,
     AttractorsViaFeaturizing,
     ClusteringConfig,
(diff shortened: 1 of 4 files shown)


KalelR:
Damn, messed up my branches! The two earlier commits actually belong to the other PR, sorry. I'll remove them.

Datseris:
use GitKraken its the best. you can start a new PR with correct commits, its cleaner.

KalelR:
> use GitKraken its the best. you can start a new PR with correct commits, its cleaner.

I know, I didn't because I thought it was expensive. Seems it's free for students though, great!

Datseris:
Its free for open source projects _in general_

PR #14, opened 2022-10-25
(WIP) Denser attractors mapped by Recurrences

KalelR:
Two additions to make the attractors found by Recurrences denser:
1. add `stop_at_dt` in step! 
2. store the attractors every time a cell if revisited, so that the number of points in attractors will be at least `mx_chk_loc_att` 

This is one fix for the problem of the algorithm finding two limit cycles when it should find only one. We found this occurs when the limit cycles stored in memory are "sparse", in the sense that there are large gaps between their points. The change No 2. does not by itself fix the problem. Something happens sometimes when `stop_at_dt=false` that makes the integrator only reach the same set of points belonging to the limit cycle and it then does not fill it densely. It seems from some quick tests that this is not only when `\Delta t` is commensurate with the cycle's period. There may be something with how it is stepping, but I don't understand that.

One point to verify is if these changes slow the code down significantly. And whether it is good to store so many points in the attractors. At the very least we should probably improve the documentation on this keyword, as it seems both George and I were confused about it (though the code itself is quite clearly written).


Datseris:
Notice that the actual value of `Δt` could be meaningless if the stop_at_dt = false. The integrators decide themselves Δt, and if the integrator is high order and can take large steps while Δt is small, then Δt is practically ignored. So, to be clear: **If the integrator step is larger than Δt, then Δt is practically entirely ignored if `stop_at_dt = false`.**

Datseris:
@KalelR please port here the example @awage used with the Moris Lecar test. We should be able to run the test once with low attractor storage and once with high, and test that once it finds two attractors and on the second time it fonds one attractor. We can make a dedicated test file specifically for the recurrences method, as we have one for cluistering. In this dedicated test file you can also test your #12

Datseris:
BTW I propose to modify the algorithm with a keyword `store_once_per_cell = true` keyword, that by default does what the current method does, and otherwise it can store arbitrarily many points per visited cell, as the current PR does.

Datseris:
I don't know where this comment is referring to.

test/mapping/recurrence_tests.jl
```diff
@@ -1,8 +1,44 @@
...
+                mx_chk_loc_att = 1000,
+                stop_at_dt,
+                store_once_per_cell,
+                sparse = true,
+                Δt = 0.1,
+                Ttr = 10,
+                )
+
+        sampler, = Attractors.statespace_sampler(Random.MersenneTwister(1);
+            min_bounds = [-0.5, -1], max_bounds = [0.5, 1])
```
awage, line 23: use these bounds: ` min_bounds = [-0.5, 0], max_bounds = [0.5, 1])`

The second variable should always be positive (for some negative initial conditions it diverges, I don't know why)

Datseris:
Alright, I need this to also progress on my end of TODOs and research in the project. So let's try to finish this ASAP. @KalelR you said:

> It seems from some quick tests that this is not only when \Delta t is commensurate with the cycle's period. There may be something with how it is stepping, but I don't understand that.

Please elaborate. What _precisely_ is going on. What were the tests you did, what was the Δt value, and what is the period of the limit cycle? Also let me know if you need something from me to do here.

Datseris:
is the system you tested with `df = Systems.morris_lecar(; I = 0.115)` ? If yes, I'll have a look and then merge.

KalelR:
> is the system you tested with `df = Systems.morris_lecar(; I = 0.115)` ? If yes, I'll have a look and then merge.

Probably not exactly, I tested for a few values in `Irange = range(0., 0.35; length = 50)`, which have several decimal places. It seemed to occur across the whole range for which the LC exists / is stable. How many points are stored in the LC does depend quite a bit on the initial condition though.

I only chose 0.115 for the unit tests, because it's a bit nicer.

Datseris:
What the actual fuck, `mapper(u0)` no longer works? What the hell did I do?!?!?! I didn't mess up with that!!!

PR #22, opened 2022-11-06
Refactor clustering into generic "grouping" interface.

Datseris:
This PR re-thinks, and in my opinion improves, the interface we have in `AttractorsViaFeaturizing`. Now, a generic "grouping" step is performed. This detaches featurizing from clustering. Trajectories are mapped into features as normal, but now, how these features are "grouped" into attractors is fleible. We have three implementations:
1. clustering
2. nearest features
3. histogram

As you will see in the source code, `AttractorsViaFeaturizing` now knows nothing about clustering. All it needs is some kind of configuration struct that implements the function `group_features`. any kind of struct that can group features therefore can be used. 

We can use the same interface in the continuation PR#10 . All we need is one more function `group_features_along_parameter` that also takes into account the fact that we have a paramater axis that may, or may not, be used.

---

The PR also re-writes the "templates" code (now nearest features) and It also improves drastically performance of the "templates" version.

---

Now the templates version and the histogram version allow the syntax `mapper(u0)` that gives you the attractor label given a **single** initial condiiton, that clustering cannot do.

---

One more thing this PR does: changes the "clustering" code to never use matrices. Vectors of static vectors are used instead. `pairwise` is super efficient in this form, and all related computations remain super efficient as well. I never understood why we make the features into a matrix. `pairwise` works perfectly fine with vectors of vectors.


KalelR (commented):
Hey, I'm a bit busy to enter into full details in the code, but it mostly seems good to me!

```
Generally speaking, the [`AttractorsViaProximity`](@ref) mapper is superior to this.
However, if the dynamical system has extremely high-dimensionality, there may be reasons
to use `AttractorsViaFeaturizing` algorithm instead, as it projects the trajectories
into a much lower dimensional representation of features.
```

I don't think this is very fair to say, especially now that we improved the `featurizing` algorithm so much. I think this referred to the supervised method of `featurizing` via clustering (in which case I agree with it) but this isn't clear in the doc.

src/mapping/attractor_mapping_featurizing.jl
```diff
@@ -1,81 +1,124 @@
...
-trajectory. See [`ClusteringConfig`](@ref) for configuring the clustering process.
+the corresponding time vector `t` and returns an `SVector{<:Real}` of features describing the
+trajectory. It is important to use static vectors for better performance.
+
+See also the intermediate functions [`extract_features`](@ref) and [`group_features`](@ref),
+which can be utilized when wanting to work directly with features.
 
 ## Keyword arguments
 * `T=100, Ttr=100, Δt=1, diffeq=NamedTuple()`: Propagated to [`trajectory`](@ref).
+* `treaded = true`: Whether to run the generation of features over threads by integrating
```
KalelR, line 51: typo, should be `threaded`

Datseris:
Tests seem to pass. They won't pass here because Entropies.jl is broken at the moment. Only after I update Entropies.jl, I'll make another PR here that finishes the Histogram method and fixes the tests (because at the moment tests will never pass no matter what).

src/mapping/attractor_mapping_featurizing.jl
```diff
@@ -106,7 +148,18 @@ function basins_fractions(mapper::AttractorsViaFeaturizing, ics::ValidICS;
...
+#####################################################################################
+# featurizing and grouping source code
+#####################################################################################
+import ProgressMeter
+
 # TODO: This functionality should be a generic parallel evolving function...
+"""
+    extract_features(mapper, ics; N = 1000, show_progress = true)
+Return a vector of the features of each initial condition in `ics` (as in
+[`basins_fractions`](@ref)), using the configuration of `mapper::AttractorsViaFeaturizin`.
```
awage, line 160:
```suggestion
[`basins_fractions`](@ref)), using the configuration of `mapper::AttractorsViaFeaturizing`.
```

src/mapping/grouping/cluster_config.jl
```diff
@@ -0,0 +1,202 @@
...
+    if rescale_features
+        features = rescale_to_01(features)
+    end
+
+    # Obtain optimal radius
+    if optimal_radius_method isa String
+        if max_used_features == 0 || max_used_features > nfeats
+            features_for_optimal = features
+        else
+            features_for_optimal = StatsBase.sample(features, max_used_features; replace = false)
```
awage, line 140: This line gave me trouble with some feature vectors. I used: 
```suggestion
          ind = sample(1:nfeats, minimum([length(features), max_used_features]); replace = false)
          features[:,ind]
```

Datseris: It's because for you `features` was a matrix. I left a comment on your PR that sample didn't work like that and you had to sample before you made them into matrix. Now, we no longer do matrices at all, so the way I sample is correct because it is done on a vector.

src/mapping/grouping/cluster_config.jl
```diff
@@ -0,0 +1,202 @@
...
+            optimal_radius_method, num_attempts_radius, silhouette_statistic
+        )
+    elseif optimal_radius_method isa Real
+        ϵ_optimal = optimal_radius_method
+    else
+        error("Specified optimal_radius_method is incorrect. Please specify the radius
+        directly as a Real number or the method to compute it as a String")
+    end
+    # Perform the DBSCAN and assign the result into labels
+    distances = pairwise(clust_distance_metric, features)
```
awage, line 153: The PR I am working on depends on these functions. For the MCBB algorithm you need to provide the distance matrix externally. I used a keyword argument `distances` wich is by default `nothing`

```
    if isnothing(distances)
        distances = pairwise(clust_distance_metric, features)
    end
```

awage:
I left only minor comments. I will need to change some things in PR #10 to adapt to the new interface. As soon as this is merged I will finish the PR.

Datseris:
Alright, I am merging this in. Some comments:

- @awage I made the source code for `group_features(..., ::GroupViaClustering)` as modular as possible. This will allow you to write your own function that calls inner functions normally, without having to pass in `distances` or do `if distances === nothing` hacks.
- @awage now that this is merged, I would recommend that you actually _close_ #10 , and open _a new pull request_ that is branching directly from `main` branch and adds your featurizing continuation. It will make the pull request much easier to review, because at the moment we have so many different parallel commits here and there that kind of inter-depend on each other. It will be much cleaner with a new PR from current `main`.
- Tests do not pass. Documentation does not build. The histogram method is not finished. However, all of these issues stem from the fact that Entropies.jl is broken at the moment. Soon I'll fix Entropies.jl and then open a _new_ pull request that implements the histogram method and fixes the tests.

awage:
The package doesn't compile due to the missing `FixedRectangularBinning`. It makes it harder to adapt the PR #10 to the new interface. Is there a quick fix to ignore this error?

Datseris:
Remove the Histogram file in grouping, or remove the type `FixedRectangularBinning` from the type. I 'll probably finish this today.

PR #29, opened 2022-11-16
Cluster attractors accross parameters

awage:
This PR is an attempt at applying the ideas of the MCCB technique to attractors instead of using features. 

TODO: 
* optimize the radius of the DBSCAN algorithm. 
* Try different distances 
* Implement tests


awage:
The clustering optimization should happen in the function `_cluster_across_parameters` of this PR.

PR #32, opened 2022-12-20
Aggregate attractors / Rework grouping-based continuation

### LINKED ISSUE #31: Grouping and matching attractors
Following is the code I wrote for grouping attractors and matching them across parameters. It contains some duplicate code from the matching algorithm. 

I applied successfully it to a population dynamics system, like seen in the figure.
populationdynamics-fractionscontinuation-fig2-method_distance_extinction-unitidx_3-samples_300-mx_chk_fnd_att_9-groupandmatch_false-th_0 5

The code is:
```julia
"""
Given the result of the continuation algorithm (`attractors_info` and `fractions_curves`), group the attractors in `attractors_info` according to the configuration established in `groupingconfig` using features defined in `featurizer`. Also, add up the basin fractions of the grouped attractors. The fractions and attractors are updated in `attractors_info` and `fractions_curves`.
Uses duplicate code from the matching algorithm.
"""
function group_and_match_continuation!(groupingconfig, attractors_info, fractions_curves, featurizer; metric = Euclidean(), threshold = Inf)
    #1. group attractors for first parameter value. 
    prev_attractors = attractors_info[1]; prev_fs = fractions_curves[1];
    group_and_relabel!(groupingconfig, prev_attractors, featurizer, prev_fs)

    #2. For subsequent parameters, group attractors and then match them with the previously grouped attractors.
    for (current_attractors, current_fs) in zip(attractors_info[2:end], fractions_curves[2:end])
        group_and_relabel!(groupingconfig, current_attractors, featurizer, current_fs)
        if !isempty(current_attractors) && !isempty(prev_attractors)
            # If there are any attractors,
            # match with previous attractors before storing anything!
            rmap = match_attractor_ids!(
                current_attractors, prev_attractors; metric, threshold
            )
            swap_dict_keys!(current_fs, rmap)
        end
        # Then do the remaining setup for storing and next step
        Attractors.overwrite_dict!(prev_attractors, current_attractors)
    end
    rmap = Attractors.retract_keys_to_consecutive(fractions_curves)
    for (da, df) in zip(attractors_info, fractions_curves)
        swap_dict_keys!(da, rmap)
        swap_dict_keys!(df, rmap)
    end
    nothing
end

"""
Group attractors based on features specified by `featurizer` and then change their labels so that
the grouped attractors have the same label.
"""
function group_and_relabel!(groupingconfig, atts, featurizer, fs=nothing)
    features = [featurizer(A) for (k, A) in atts]
    newlabels = length(features) > 1 ? group_features(features, groupingconfig) : keys(atts)
    rmap = Dict(keys(atts) .=> newlabels)
    swap_dict_keys!(atts, rmap)
    !isnothing(fs) && sum_fractions_keys!(fs, rmap)
    return rmap
end

"""
Given a replacement map that is a dictionary mapping the old keys in `fs` to new keys,
update `fs` to the sum of the values of `fs` with the same key.
"""
function sum_fractions_keys!(fs, rmap)
    newkeys = unique(values(rmap))
    fssum = Dict(newkeys .=> 0.0)
    for oldkey in keys(fs)
        newkey = rmap[oldkey]
        fssum[newkey] += fs[oldkey]
        pop!(fs, oldkey)
    end
    for newkey in newkeys
        fs[newkey] = fssum[newkey]
    end
    return nothing
end
```

Additionally, there is a small fix needed in the `group_features` function (line 131 of `cluster_config.jl`). The line needs to be 
```
    distances = pairwise(config.clust_distance_metric, features) 
```

Currrently, the order of the arguments is inverted and it is trying to access `config.metric` which does not exist. 

The full code I used to generate the figures is 
```julia
using DrWatson
@quickactivate
using OrdinaryDiffEq:Vern9
using DynamicalSystemsBase
using Attractors, Random
using CairoMakie
using Colors
include("$(srcdir())/vis/basins_plotting.jl")

monod(r, R, K) = r*R/(K+R)
function μ!(μs, rs, Rs, Ks)
    for i in eachindex(μs)
        mo1 = monod(rs[i], Rs[1], Ks[1,i])
        mo2 = monod(rs[i], Rs[2], Ks[2,i])
        mo3 = monod(rs[i], Rs[3], Ks[3,i])
        μs[i] = min(mo1, mo2, mo3)
    end
    nothing
end
#not the most optimied but w/e
function Rcoup!(Rcoups, Ns, Rs, μs, cs)
    fill!(Rcoups, 0.0)
    for j in eachindex(Rcoups)
        for i in eachindex(μs)
            Rcoups[j] += cs[j,i] * μs[i] * Ns[i]
        end
    end
    nothing
end

function competition!(du, u, p, t)
    @unpack rs, Ks, ms, Ss, cs, μs, Rcoups, D = p
    n = size(Ks, 2)
    Ns = view(u, 1:n)
    Rs = view(u, n+1:n+3)
    dNs = view(du, 1:n)
    dRs = view(du, n+1:n+3)
    μ!(μs, rs, Rs, Ks)
    Rcoup!(Rcoups, Ns, Rs, μs, cs)
    @. dNs = Ns * (μs - ms)
    @. dRs = D*(Ss - Rs) - Rcoups
    nothing
end

mutable struct CompetitionDynamics
    rs :: Vector{Float64}
    ms :: Vector{Float64}
    Ss :: Vector{Float64}
    μs :: Vector{Float64}
    Rcoups :: Vector{Float64}
    Ks :: Matrix{Float64}
    cs :: Matrix{Float64}
    D :: Float64
end

function CompetitionDynamics(fig="1")
    if fig == "4" || fig == "1"
        Ks  = [
            0.20 0.05 0.50 0.05 0.50 0.03 0.51 0.51;
            0.15 0.06 0.05 0.50 0.30 0.18 0.04 0.31;
            0.15 0.50 0.30 0.06 0.05 0.18 0.31 0.04;
        ]

        cs = [
            0.20 0.10 0.10 0.10 0.10 0.22 0.10 0.10;
            0.10 0.20 0.10 0.10 0.20 0.10 0.22 0.10;
            0.10 0.10 0.20 0.20 0.10 0.10 0.10 0.22;
        ]
        if fig == "1"
            Ks = Ks[:, 1:5]
            cs = cs[:, 1:5]
        end
    elseif fig == "2" || fig == "3"
        Ks = [
            0.20 0.05 1.00 0.05 1.20;
            0.25 0.10 0.05 1.00 0.40;
            0.15 0.95 0.35 0.10 0.05;
        ]

        cs = [
            0.20 0.10 0.10 0.10 0.10;
            0.10 0.20 0.10 0.10 0.20;
            0.10 0.10 0.20 0.20 0.10;
        ]

    else
        @error "nope"
    end

    N = size(Ks, 2)

    rs = [1.0 for i=1:N]
    D = 0.25
    ms = [D for i=1:N]
    Ss = [10.0 for j=1:3]
    μs = zeros(Float64, N)
    Rcoups = zeros(Float64, 3)
    return CompetitionDynamics(rs, ms, Ss, μs, Rcoups, Ks, cs, D)
end


"""
Given the result of the continuation algorithm (`attractors_info` and `fractions_curves`), group the attractors in `attractors_info` according to the configuration established in `groupingconfig` using features defined in `featurizer`. Also, add up the basin fractions of the grouped attractors. The fractions and attractors are updated in `attractors_info` and `fractions_curves`.
Uses duplicate code from the matching algorithm.
"""
function group_and_match_continuation!(groupingconfig, attractors_info, fractions_curves, featurizer; metric = Euclidean(), threshold = Inf)
    #1. group attractors for first parameter value.
    prev_attractors = attractors_info[1]; prev_fs = fractions_curves[1];
    group_and_relabel!(groupingconfig, prev_attractors, featurizer, prev_fs)

    #2. For subsequent parameters, group attractors and then match them with the previously grouped attractors.
    for (current_attractors, current_fs) in zip(attractors_info[2:end], fractions_curves[2:end])
        group_and_relabel!(groupingconfig, current_attractors, featurizer, current_fs)
        if !isempty(current_attractors) && !isempty(prev_attractors)
            # If there are any attractors,
            # match with previous attractors before storing anything!
            rmap = match_attractor_ids!(
                current_attractors, prev_attractors; metric, threshold
            )
            swap_dict_keys!(current_fs, rmap)
        end
        # Then do the remaining setup for storing and next step
        Attractors.overwrite_dict!(prev_attractors, current_attractors)
    end
    rmap = Attractors.retract_keys_to_consecutive(fractions_curves)
    for (da, df) in zip(attractors_info, fractions_curves)
        swap_dict_keys!(da, rmap)
        swap_dict_keys!(df, rmap)
    end
    nothing
end

"""
Group attractors based on features specified by `featurizer` and then change their labels so that
the grouped attractors have the same label.
"""
function group_and_relabel!(groupingconfig, atts, featurizer, fs=nothing)
    features = [featurizer(A) for (k, A) in atts]
    newlabels = length(features) > 1 ? group_features(features, groupingconfig) : keys(atts)
    rmap = Dict(keys(atts) .=> newlabels)
    swap_dict_keys!(atts, rmap)
    !isnothing(fs) && sum_fractions_keys!(fs, rmap)
    return rmap
end

"""
Given a replacement map that is a dictionary mapping the old keys in `fs` to new keys,
update `fs` to the sum of the values of `fs` with the same key.
"""
function sum_fractions_keys!(fs, rmap)
    newkeys = unique(values(rmap))
    fssum = Dict(newkeys .=> 0.0)
    for oldkey in keys(fs)
        newkey = rmap[oldkey]
        fssum[newkey] += fs[oldkey]
        pop!(fs, oldkey)
    end
    for newkey in newkeys
        fs[newkey] = fssum[newkey]
    end
    return nothing
end

reduced_grid(grid, newlength) = map(g -> range(minimum(g), maximum(g); length = newlength), grid)

figidx = "2"
p = CompetitionDynamics(figidx)

N = size(p.Ks, 2)
u0 = [[0.1 for i=1:N]; [S for S in p.Ss]]
ds = ContinuousDynamicalSystem(competition!, u0, p, (J, z, p, t)->nothing)
diffeq = (alg = Vern9(), maxiters=Inf);
int = integrator(ds, u0; diffeq)


function _default_seeding_process_deterministic(attractor::AbstractDataset)
    max_possible_seeds = 10
    seeds = round(Int, log(10, length(attractor)))
    seeds = clamp(seeds, 1, max_possible_seeds)
    return (attractor.data[i] for i in 1:seeds)
end

isextinct(A, idx) = all(A[:, idx] .<= 1e-2)
function get_metric(unitidx=nothing)
    if isnothing(unitidx)
    return "euclidean", Euclidean();
    else
        distance_extinction = function(A,B, idx)
            A_extinct = isextinct(A,idx)
            B_extinct = isextinct(B,idx)
            return (A_extinct == B_extinct) ? 0 : 1
        end
        return "distance_extinction", (A, B) -> distance_extinction(A, B, unitidx);
    end
    nothing
end


pidx = :D; ps = 0.2:0.005:0.3;
xg = range(0, 60,length = 300); grid = ntuple(x->xg, N+3);
unitidxs = [3]
samples_per_parameter = 300
mx_chk_fnd_att = 9
groupandmatch_bools = [false, true]; groupandmatch_savetogether = true
threshold = 0.5
for unitidx in unitidxs
    info_extraction = A -> isextinct(A, unitidx)
    metricname, metric = get_metric(unitidx)
    ds = ContinuousDynamicalSystem(competition!, u0, p, (J, z, p, t)->nothing);
    mapper = AttractorsViaRecurrences(ds, grid;
            Δt= 1.0,
            mx_chk_fnd_att,
            diffeq,
        );
    continuation = RecurrencesSeedingContinuation(mapper; seeds_from_attractor=_default_seeding_process_deterministic, metric, threshold, info_extraction);
    # continuation = RecurrencesSeedingContinuation(mapper; metric, threshold, info_extraction);
    sampler, = statespace_sampler(Random.MersenneTwister(1234);
        min_bounds = minimum.(grid), max_bounds = maximum.(grid)
    );
    fractions_curves_og, attractors_info_og = basins_fractions_continuation(
        continuation, ps, pidx, sampler;
        show_progress = true, samples_per_parameter
    );
    fig = nothing
    for (idx, groupandmatch) in enumerate(groupandmatch_bools)
        fractions_curves = deepcopy(fractions_curves_og)
        attractors_info = deepcopy(attractors_info_og)
        if groupandmatch
            featurizer(A) = [Int32(A)]
            groupingconfig = GroupViaClustering(; min_neighbors=1, optimal_radius_method=0.5) #note that minneighbors = 1 is crucial for grouping single attractors
            metric = function distance_function_bool(A,B)
                return abs(A - B)
            end
            group_and_match_continuation!(groupingconfig, attractors_info, fractions_curves, featurizer; metric=distance_function_bool)
        end

        #plot details
        ukeys = unique_keys(attractors_info)
        label_extincts = map(atts->[k for (k,v) in atts if v == 1], attractors_info); label_extincts = unique(vcat(label_extincts...))
        label_surviving = [key for key in ukeys if key ∉ label_extincts]
        legend_labels = [label in label_extincts ? "extinct" : "surviving" for label in unique_keys(attractors_info)]
        colors_surviving = length(label_surviving) == 1 ? [colorant"green"] : collect(range(colorant"darkolivegreen2", stop=colorant"green", length=length(label_surviving)))
        colors_extinct   = length(label_extincts) == 1 ? [colorant"red"] : collect(range(colorant"red", stop=colorant"red4", length=length(label_extincts)))
        colors_bands = [colorant"white" for _ in unique_keys(attractors_info)]
        colors_bands = merge(Dict(label_surviving .=> colors_surviving), Dict(label_extincts .=> colors_extinct))
        #plot
        idxrow = groupandmatch_savetogether ? idx : 1
        fig, ax = basins_fractions_plot(fractions_curves, collect(ps); idxrow, fig, add_legend=true, legend_labels, colors_bands);
        ax.xlabel = "D";
        ax.ylabel = "fractions";
        ax.title = ["not grouped", "grouped"][idxrow]
        vlines!(ax, 0.25, color=:black);
        !groupandmatch_savetogether && save("$(plotsdir())/defaultseeding-populationdynamics-fractionscontinuation-fig$figidx-method_$metricname-unitidx_$(unitidx)-samples_$(samples_per_parameter)-mx_chk_fnd_att_$(mx_chk_fnd_att)-groupandmatch_both-th_$(threshold).png", fig, px_per_unit=3)
        # fig
    end
    groupandmatch_savetogether && save("$(plotsdir())/defaultseeding-populationdynamics-fractionscontinuation-fig$figidx-method_$metricname-unitidx_$(unitidx)-samples_$(samples_per_parameter)-mx_chk_fnd_att_$(mx_chk_fnd_att)-groupandmatch_$groupandmatch-th_$(threshold).png", fig, px_per_unit=3)
end



```

Datseris:
_(because this PR is very large, the git diff is not very useful. Manually checkout and read the docstrings for the continuation and aggregation functions)_

# Motivation 

Alright this is a really big PR that implements what I envisioned as the API for aggregating already found attractors from our recurrences algorithm. The gist is as follows: all three of us (@Datseris, @awage, @KalelR ) have coded some kind of version that aggregates attractors. But I wanted to be greedy. I wanted the best possible infrastracture. I wanted it all.

I wanted us to be able to use **any grouping configuration** to aggregate attractors. Histograms, nearest feature, clustering. I also wanted us to be able to all these grouping configurations for the "traditional" featurizing based continuation. I.e., what MCBB does, but for any possible way to group, not only clustering. And I wanted all of this to be as simple as possible with the least amount of code duplication. This is what this PR does.

Closes #31
Closes #29

# Changes done

- Reworked the code of `GroupViaClustering`:
  - It now has the `mmap` keyword by itself, and uses mmap even for the non-continuation if the user wants.
  - It now has a "cheating" way of either including the parameter weight of the distance matrix or not when creating the distance matrix. This cheating way is not part of the public API.
- Rewritten the clustering across parameters continuation pretty much from scratch. It is now `GroupAcrossParameterContinuation`. This reflects the design I always wanted: that the code would work for any grouping method. Amazingly, even though the code is now much more general, it actually is less lines of code. This is due to the redesign of the clustering struct.
- Created a new `aggregate_attractor_fractions` that can aggregate the fractions of the output of `RecurrencesSeededContinuation` using the grouping pipeline of `AttractorsViaFeaturizing`.

# TODOs

- [ ] Tests for new features
- [x] Fix tests of the adjusted Grouping continuation
- [ ] Tests that all grouping configurations work with the Grouping continuation method
- [x] Update docstring of Grouping continuation and of `GroupingConfig`.
- [x] Finish docstring of `aggregate_attractor_fractions`.
- [ ] Add a new documentation section for the new aggregation function
- [ ] Add Kalel competition dynamics in the a doc section about the aggregation function.
- [x] Remove label `-1` from aggregated fractions if it is always with 0 fraction. 
- [x] Return information about aggregated features, not only the aggregated fractions

# Example

Here is how a full script of using this new functionality would look for the competition dynamics example of @KalelR . Pay attention to how simple the script actually is, in comparison to the code Kalel has in #31. For me, this is the biggest advantage of this PR. That the final API is not only super simple, but re-uses existing knowledge. As long as you know the `GroupVia...` interface, you can immediatelly use it to aggregate the output of `RecurrencesSeededContinuation`.

```julia
using DrWatson
@quickactivate
using Attractors, OrdinaryDiffEq, CairoMakie
using Random
include(srcdir("vis", "basins_plotting.jl"))
include(srcdir("fractions_produce_or_load.jl"))
include(srcdir("additional_predefined_systems.jl"))

N = samples_per_parameter = 300
P = total_parameter_values = 21
fig, axs = subplotgrid(2,1)
axs[1].title = "original"
axs[2].title = "grouped"
display(fig)

# Population dynamics recurrences continuation. It uses the DrWatson integration
# pipeline to not re-compute things. But all in all, it is a direct call to
# `RecurrencesSeededContinuation` with default matching behavior
# (distance in state space). **No special matching or metric is used here!!!**
ds = competition()
mapper_config = (; Δt= 1.0, mx_chk_fnd_att=9);
xg = range(0, 60; length = 300);
grid = ntuple(x->xg, 8);
pidx = :D
prange = range(0.2, 0.3; length = P)
config = FractionsRecurrencesConfig("populationdynamics", ds, prange, pidx, grid, mapper_config, N)

output = fractions_produce_or_load(config; force = false)

@unpack fractions_curves, attractors_info = output

basins_fractions_plot!(axs[1,1], fractions_curves, prange)

# Aggregation of attractors based on the existence or not of some unit
unitidxs = 3

featurizer = (A) -> begin
    i = isextinct(A, unitidxs)
    return SVector(Int32(i))
end
isextinct(A, idx) = all(a -> a <= 1e-2, A[:, idx])

# `minneighbors = 1` is crucial for grouping single attractors
groupingconfig = GroupViaClustering(; min_neighbors=1, optimal_radius_method=0.5)

aggregated_fractions, aggregated_info = aggregate_attractor_fractions(
    fractions_curves, attractors_info, featurizer, groupingconfig
)

basins_fractions_plot!(axs[2,1], joint_fractions, prange)

wsave(plotsdir("gd", "competition_dynamics_aggregation.pdf"), fig)
```

which produces

image


Datseris:
@KalelR notice that the figure I get is different form the one you get in #31 . But this happens because you do not use the "pure" version of `RecurrencesSeededContinuation`. You alter the matching that happens in the algorithm. I do not know whether this is important or not, or why you do it. I thought we want to get teh "pure" attractors first, and then aggregate them based on whether some species is alive or not.

KalelR:
> @KalelR notice that the figure I get is different form the one you get in #31 . But this happens because you do not use the "pure" version of `RecurrencesSeededContinuation`. You alter the matching that happens in the algorithm. I do not know whether this is important or not, or why you do it. I thought we want to get teh "pure" attractors first, and then aggregate them based on whether some species is alive or not.

Yup, that is fine. If I'm not mistaken, I used the extinction criteria both for the continuation algorithm and for the grouping later, which isn't needed. But would that lead to a difference in the grouped fractions? One point to consider for this system is that in that intermediate parameter range (D ~ 0.25) the basin boundaries are fractal, so changing any seed would change the results a bit.

awage:
Hi, I have done a bit of thinking on the interface and I think we can go further in the simplification. I put you a schematics of what is my idea: 

workflow

Most of the pieces are already in place but the interface is still too intricate. I believe we can separate the functionallities more clearly and provide general interface for commodity. For example `RecurrencesSeededContinuation` can be separated in two functionalities: the seeding and the grouping after you get the labels. 

The idea is to distinct clearly the three stages: sampling -> grouping -> fractions labelling.  The last stage should be quite slim once the grouping has been done.

Datseris:
I am trying to figure out what you are suggesting, or what you are asking for, but I can't :( I will probably need a more detailed explanation. I had a look at the attached figure.

---

> For example `RecurrencesSeededContinuation` can be separated in two functionalities: the seeding and the grouping after you get the labels.

I think there is some confusion on the word "seeding". The reason we have "Seeding" in `RecurrencesSeededContinuation` is because at the new parameter some initial conditions are seeded from the previous attractors. This targets of overcomning the bottleneck of finding the new attractor in the new parameter. This seeding has nothing to do with sampling the state space.

Besides, I don't see any way to make any further separations in the recurrences continuiation method. Attractors need to stay in memory because you need them to match the attractors in the new parameter. Initial conditions are never stored in memroy, only their overall fractions are.

After you have finished finding the attractionrs and their fractions, you can still "group" them using the new function `aggregate_attractor_fractions`.

---

> (in the figure:) For each initial condition and each parmaeter value you get a label

This is already possible for all mappers except the mapper `AttractorsViaFeaturizing` with `GroupViaClustering`. You can do this already, there is no re-design necessary. E.g.,
```julia
ds = some dynamical system
group_confg = GroupByHistogram(...)
p = parameter value
mapper = AttractorsViaFeaturizing(featurizer, ds, group_config)
# Do what you ask for:
set_parameter!(ds, 1, p)
u0 = random initial condition
id = mapper(u0)
```

And you can still group across any parameter range you want by simply calling the `aggregate_attractor_fractions` with `identity` as the featurizer.

---

> The idea is to distinct clearly the three stages: sampling -> grouping -> fractions labelling. The last stage should be quite slim once the grouping has been done.

I do not understand what "fractions labelling" means here. Actually, I don't know what "sampling" means either. You sample an initial condition at a given parameter. Sure. That's trivial. What do you do with it though I don't get it. How can you group initial conditions before doing anything with them? You could make them into features, and then group them. But that's alreay what we do. This is already two distinct phases. The fractions labelling is a no-code phase: you just label the different groups with the integers. So that's already a different stage...?

---

> image

I don't understand what these three boxes mean or why they are parallel to each other, as they are different things. And why does one box has both MCBB and DBSCAN, and the other only DBSCAN. For you, what is "MCBB"?

---

To make things easier @awage please  try to state as precisely as possible exactly what you would like to do. In the practical sense; what you would like to do e.g., in your script. Being precise will help me understand what you want and whether it really cannot be done with the current code base. Because so far I dont' see something you want to do that isn't already possible.

awage:
I will try to make a better exposition of my ideas. But maybe the problem is my side and I don't see the full picture. So after all it is to make sure that we are on the same page regarding the design. Tomorrow I will try to make my point more precisely.

awage:
Hi, sorry I though I had posted my comment yesterday but I didn't click the green button. I leave you some slides with my view of the design. I want to open this discussion since I think it worth having a clear and simple design. And furthermore the data oriented design allows to leverage the multiple dispatch assets of Julia. 

Please let me know what is your point of view. Maybe it is still too vague and you want more specific comments.

[pres_design_george_kalel.pdf](https://github.com/JuliaDynamics/Attractors.jl/files/10296957/pres_design_george_kalel.pdf)

Datseris:
As far as I can tell, everything on the slides is already happening with current design. If you can give me a specific example of something that you have on the slides that is not part of current design please go ahead. but from my point of view, current functionality is as general as you would like it on the slides.

---

The first column on this table doesn't make sense: 

image

 MCBB is literally the Featurizing approach with DBSCAN grouping algorithm.

awage:
I agree all functionalities are already coded. It is more a question of code organization. 

So to give you a concrete example,  the function `GroupAccrossParameterContinuation` functionalities could have also a `AttractorsViaRecurrence` mapper as an input instead of defining the `aggregate_attractors_fractions`. 

The same way, `RecurrencesSeedingContinuation` could be just `SeedingContinuation` or `ParametricContinuation` so that we can also match features from one parameter slice to another.  

This is also the meaning of the column MCBB, what I mean is clustering using the DBSCAN and the parameter weights.  

All my comments are in the sense of slimming down the interface but opening the possibilities of combination between mappers and "continuators".

Datseris:
> `GroupAccrossParameterContinuation` functionalities could have also a `AttractorsViaRecurrence` mapper as an input instead of defining the `aggregate_attractors_fractions`

The function `aggregate_attractors_fractions` is useful because it can also be used to aggregate the results of a single parameter application. I.e., if `mapper = AttractoprsRecurrences()` and `fracs, attractors = basins_fractions(mapper)`, you can still aggregate this result. Which makes sense and is useful outside the continuation point. It is also useful in case you want to keep both the attractors but an aggregated result. So you want to keep both the result of `RecurrencesSeededContinuation`, but also aggregate them to present a nice plot. In this case, you need this function `aggregate_attractors_fractions`. How would your alternative deal with this case, since you propose that it replaces ``aggregate_attractors_fractions`?

Additionally, as it is right now `GroupAccrossParameterContinuation` takes in a AttractorsViaFeaturizer mapper, and this mapper has a featurizing method. If we gave it the `AttractorsViaRecurrence`, when is the featurizer coming from? The recurrences mapper doesn't have a featurizer, so the featurizer would need to be an additional argument. So you actually need to make a different function anyways, and you don't actually slim down anything when compared to calling `aggregate_attractors_fractions`.

---

> `RecurrencesSeedingContinuation` could be just `SeedingContinuation` or `ParametricContinuation` so that we can also match features from one parameter slice to another.

But what does this mean though? What is a `SeedingContinuation`? You can only seed if you have attractors to seed from. And the only method we have so far that finds attractors is the Recurrences one. So what does it even mean to have a method `SeedingContinuation`. What does this method do? _Exactly?_ 

I really don't understand what you mean when you say "seeding". I use this word exclusively for the recurrence method that seeds initial conditions from attractors to find the next attractors fast. This has nothing to do with sampling random initial conditions so that we calculate the fractions. For me seeding as a word only makes sense to use for a method that finds the _actual_ attractors, not features. The only method we have that does this is the recurrences one.

---

> `ParametricContinuation`

What does this mean, exactly? All continuation methods are "parametric", since they continue along a parameter axis...

awage:
> How would your alternative deal with this case, since you propose that it replaces ``aggregate_attractors_fractions`?

It can a low level method for `GroupAccrossParameterContinuation`

> If we gave it the AttractorsViaRecurrence, when is the featurizer coming from? The recurrences mapper doesn't have a featurizer, so the featurizer would need to be an additional argument

You just have to define a special distance for the DBSCAN, that's all. 

> But what does this mean though? What is a SeedingContinuation? You can only seed if you have attractors to seed from. And the only method we have so far that finds attractors is the Recurrences one. So what does it even mean to have a method SeedingContinuation. What does this method do? Exactly?

This is my very bad naming habit. Everything boils down to two alternatives to do the continuation: 
    * You collect the metrics for different ICs and different parameters and try to match them with a grouping method, for example DBSCAN. This is `GroupAccrossParameterContinuation`.
    * You compute the metric \rho(p,x) for the parameter p and try to match it with \rho(p+\delta p, x). This is the method used in `RecurrencesSeedingContinuation`. 

I don't see any reason why we cannot use the different inputs for these two methods. I mean that it doesn't matter if the input is a vector of `Features` or a vector of `Datasets` with attractors in it. In the end, the attractors can be considered as special features. The seeding method can be an option for the `RecurrencesSeedingContinuation`. So it would leave us only with the `Continuation`  word.

Datseris:
Unfortunately I still don't get what you want. The bullet points in the comment do not make sense. In `RecurrencesSeededContinuation` we we go from p to p+δp because this is the change in parameter where the attractosr will have changed the least, and hence we will find them the fastests and the matching of IDs will be the most accurate. The `GroupingAcrossParameterContinuation` does not have this matching accuracy, because it matches by features. So for it it doesn't matter at which parameters we collect features at. 

Since it is still not clear to me how to proceed, here's what I'll do: I'll finish this PR by finishing the tests, adding documentation, and then merging in. Since this PR provides all functionality we need for our paper, I want to have it in.

Then, to improve things in the direction you want, there are two options: we need to have a live video call so we talk in real time and hopefully understand what you want to happen. Or, you open a pull request showing the changes you want to do and then I see concrete code, where there is no ambiguity what is happening.

awage:
> Since it is still not clear to me how to proceed, here's what I'll do: I'll finish this PR by finishing the tests, adding documentation, and then merging in. Since this PR provides all functionality we need for our paper, I want to have it in.

This is wiser. I will propose a PR with the changes that I have in mind and then we can discuss it is worth moving in this direction. In the meantime if you need a hand with one of the TODOs or with the text please tell me.

PR #40, opened 2023-02-12
Change the interface for grouping and matching attractors 

awage:
This PR is part of an effort to bring the same continuation capabilities to Features mapper and Recurences mapper. 

This is the part 1. What has been done so far: 

* Complete rewrite of the matching routine. It is mostly moving the code around in order to separate the two important steps: collecting the attractors and then the matching. 
* Adding a primitive form of grouping attractors. A distance matrix is computed and passed to a low level function of dbscan. This must be improved in order to pull the muscle of the grouping capabilities of `Attractors.jl`. 
* Very basic test.

There are some issues about using attractors as Features. I haven't solve the problem of the clustering with an automatic epsilon radius. Maybe it is not necessary.  Also I have to test with other function info extraction (such as the mean of the Datasets). 
  
It is still primitive but it will let you know where I want to get to. The code is not ready for review. There are still plenty of details that I need to take care of. 

The part 2 of the PR has not started yet, it will be adding the slice matching to the Feature mapper.


awage:
The part 1 and 2 of the changes I am proposing are almost done. The outline of the changes are: 
* There are now two ways of doing continuation: with recurrences and featuring. 
* You can do both matching between slices and grouping accross parameters. 

I don't think the interface is the ideal. Maybe breaking the matching and grouping into separate functions is better. We can discuss that. 

The grouping of the attractors does not use the `group_features` function because of the type of `attractors` . These are datasets and features are vectors. Now the grouping is done by computing the distance between attractors pairwise and fed to the DBSCAN algorithm with a fixed threshold. I don't think it is a big issue since the number of attractors is fairly small. Also it makes sense to set a fixed threshold as in the matching process. In this aspect, the functions in `aggregator` become obsolete since the grouping is doing the same job. Now you can also group with a feature function on the attractors. 

TODOs: 
* Documentation 
* Try it on "real" cases such as the kuramoto net. 
* Reduce the codebase by sharing functions between the Featuring and Recurrence part. There are some progress to be done there.

Datseris:
Why are there some many file conflicts...? Where did you branch of to do this? I mentioned in the email to branch of the latest version of code I've pushed to include all changes from DynamicalSystems.jl 3.

Datseris:
(also, remember, it is "Featurizing" not "Featuring" :D ) 

Let me know when the docs are done and I will review.

awage:
It's a matter of depencencies for the tests. I was not abble to run the code on the latest version. I will merge the latest commits into the PR so you can checkout the branch cleanly.

awage:
> (also, remember, it is "Featurizing" not "Featuring" :D )
> 
> Let me know when the docs are done and I will review.

Damn... you're right

awage:
I'll start working on the docs.

awage:
@Datseris I have finished for now the part 1 and 2 of my program. I left the part 3 for later (the stuff about using the parameter p as an extra feature).

Lots of things have changed but the basic functionalities of continuation still remain. I believe that the code is more general now and more structured. I still have doubt about deleting the `aggregate_attractors` file. We can discuss the details once you have seen the code.  I have been using the code on some examples but I still need to perform more profound tests. 

The funny thing is that some piece of code you have programmed for attractors work perfectly well with features ( `match_attractors_id!`). So nice!! 

It is harder to use the code for grouping features with attractors since they are dictionaries of `Datasets`. They are not compatible with the interface of `group_features`. Maybe it is possible to adapt an interface but I could not do it.

src/continuation/continuation_featurizing.jl
```diff
@@ -0,0 +1,205 @@
...
+
+- Features are grouped across a parameter range. Said differently, all features
+of all initial conditions across all parameter values are put into the same "pool"
+and then grouped as dictated by the `group_config` of the mapper.
+After the grouping is finished the feature label fractions are distributed
+to each parameter value they came from.
+- The features are grouped and labeled for a single parameter slice. The continuation
+is performed by matching the labelled features from one parameter slice to the next 
+using the special distance `method`.  
+- The labelled features are grouped using an histogram binning. TODO
```
Datseris, line 32: _How_ do I choose which method is used...?

awage: The method is selected in `basins_fractions_continuation`. This is a design point I am not completely satisfied with.

Datseris: Okay I have a solution, we talk in 10!

Datseris: taking notes on the code now...

awage:
Close this PR until further advance in the design.

PR #46, opened 2023-03-01
Always return attractors

Datseris:
I am preparing a tutorial on Attractors.jl and I found itr very annoying that `basins_fractions` won't return the attractors if a sampler is given. This is changed here so that attractors are always returned.

Technically this is a brteaking change, but I Am assuming this package doesn't have a user base yet so I'll go ahead with it...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# v1.1
+- `basins_fractions` with a mapper will always return attractors.
+
+# v1
+- Added the `continuation` function and the two types `RecurrencesSeededContinuation` and `GroupAcrossParametersContinuation`
+
 # v0.1.0
 This is the first release. It continues from ChaosTools.jl v2.9, and hence, the comparison of attractor-related features is w.r.t to that version.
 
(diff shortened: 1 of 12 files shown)


awage:
This change simplifies some code of mine also.

Datseris:
The problem is, this change is not possible. It is not possible for the clustering method to return the attractors if `ics` is not a state space set but a function. So I have to think this through a bit...

awage:
The problem is for the Featurizing mapper I guess? In this case, you have to change the `extract_attractors` function and dispatch on the mapper type.

Datseris:
I have a solution; i am implementing it now

Datseris:
Okay, the solution was to make the function
```
```
part of the API and make it work for the featurizing mapper as well, but only when `ics` is a statespaceset. For other mappers it always works.

I'll put this in the examples as well and merge and tag.

Datseris:
Yes, I also though of deepcopying the original sampler into the `AttractorsViaFeaturizing`. This can be done as a PR in the future!

Issue #55, opened 2023-03-30
Function `_rescale_to_01` is incorrect


KalelR:
Hey 

I just looked at the `_rescale_to_01` function, which we use to rescale each feature dimension onto the `[0,1]` interval, and noticed it is clearly incorrect. 

Currently, it's 

```
function _rescale_to_01(features::AbstractStateSpaceSet)
    mini, maxi = minmaxima(features)
    return map(f -> f .* (maxi .- mini) .+ mini, features)
end
```

This does not rescale features to `[0,1]`. 

Instead, it should be 
```
function _rescale_to_01(features::AbstractStateSpaceSet)
    mini, maxi = minmaxima(features)
    return map(f -> (f .- mini) ./ (maxi .- mini), features)
end
```

When I first wrote it, it was correct, I'm pretty sure. Somewhere in the middle of the refactoring we got this wrong. It might very well nott affect results, but we should correct it to do what it should. 

I'm writing an issue because I implemented this and the dummy bistable test in `grouping_continuation.jl` got an error. I didn't have time to understand why this happens.. but I saw that this rescaling of features leads to small floating point errors (e.g. the rescaled features being `0.49999999...2` instead of `0.5`). Maybe the matching procedure does not match them because of the errors, thus creating new attractors? Or is the rescaling applied to the global pool (which would rescale the features a certain way) and then also applied to a local pool, for only attractor (whihc would rescale in a different way)?

Datseris:
Well sure why not do the PR that fixes the function?

Datseris:
seems to me you are correct and there isn't much debate to be had here :D

Datseris:
bump

KalelR:
The dummy bistable map in `grouping_continuation.jl` tests fails with this change. It apparently leads to floating point errors that somehow screw up the matching. The reason for this is not obvious to me, but I thought could be for you. If not, I can take a look at this sometime in the next days :)

Datseris:
Can you open a PR that does this, and then I look at the failing test?

PR #59, opened 2023-04-06
Fix `_rescale_to_01`

KalelR:
As discussed in Issue #55

### DIFF
--- a/test/continuation/grouping_continuation.jl
+++ b/test/continuation/grouping_continuation.jl
@@ -36,7 +36,7 @@ using Random
     featurizer(a, t) = a[end]
     clusterspecs = Attractors.GroupViaClustering(optimal_radius_method = "silhouettes", max_used_features = 200)
     mapper = Attractors.AttractorsViaFeaturizing(ds, featurizer, clusterspecs; T = 20, threaded = true)
-    continuation = GroupAcrossParameterContinuation(mapper; par_weight = 0.1)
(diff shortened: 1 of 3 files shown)


awage:
Do you want me to have a look at this?

Datseris:
yes please!

awage:
I fixed the test but I don't understand what is happening. The `par_weight` add an extra weight to the features far away in parameter space so they don't get clustered together. For the example in the dumb map, if I set this value to 0.1 the test fails because the distance between clusters is too big. For 0.09 it works. I have no clue.

test/continuation/grouping_continuation.jl
```diff
@@ -49,7 +48,7 @@ using Random
             k = sort!(collect(keys(fs)))
             @test sort!(collect(keys(infos))) == k
             @test length(k) == 1
-            @test infos[1] == [0, 0]
+            # @test infos[1] == [0, 0]
```
Datseris, line 51: Why is this commented out? The way to fix tests isn't to comment them out :D

awage: Yeah! that was too easy!! It is not the line that was giving trouble. I will check this.

Datseris:
all tests pass so this seems okay? in the end the problem was wrongly giving parameter weight to the simulation?

awage:
Changed version number.

awage:
Wait, there is a version mismatch between main branch and this PR. Main is in 1.3.0 and this PR is in 1.2.5 

I update to 1.3.1

Issue #60, opened 2023-04-19
Octopus like basins


Datseris:
I've just read Basins with Tentacles by Yuanzhao Zhang and Steven H. Strogatz:

![image](https://user-images.githubusercontent.com/19669089/233044241-817bb0d9-cc18-4827-ab29-efaa3eafba9b.png)

The authors say that a basin is octopuslike if, starting from the attrractor, a random perturbation in a given direction exits and re-enters the basin later when increasing the strength of the perturbation, and this continues many times.

I'm opening a feature request here, mainly because this is tightly linked with the minimal fatal shock algorithm #4 . We can kill two birds with one stone and reuse the same code. 

In essense, we write a code that produces perturbations along directions. This code produces the result of a boolean matrix of dimensions as much as teh state space (1 = radious = perutrbation strenght, the other dimensions are the anglers). 

By careful analysis of this array we can find both (1) the minimal perturbation along a direction, and (2) whether there is octopus like structure in this direction. The minimal fatal shock #4 is the minimum of (1) and the basin is octopuslike if all directions are octopuslike from (2).

awage:
This seems a very hard problem in high dim. I have the project to implement some ideas from the following paper:
https://arxiv.org/abs/2207.08241

They suggest to use techniques from statistical physics to sample volume of basins in high-dim  (see last section). By the way they cite Strogatz paper. There are also MCMC techniques suitable for this kind of sampling. I want to experiment with this stuff this summer. Very recommended reading. 

There is not much done in the dynamical systems community for this kind of system. Strogatz paper is one of the few. But there is a very large corpus of research in statistical physics and computational chemistry.

Datseris:
Yes, the strogatz paper cites this: 

![image](https://user-images.githubusercontent.com/19669089/233090265-926b6f07-bc7e-42ac-bceb-db3a7987689c.png)


Which I thought sounds super promising to add here, but after reading it a bit I realized that it really only applies to statistical physics concept where you have some sorts of free energy that you can minimize.

awage:
Stefano Martiniani is one of the authors of the arxiv paper I sent. The arxiv version is more understandable and they try to bridge the gap between dynamical systems and problems like hard sphere packing but they don't offer examples. Also I missed an implementation of the technique. There is no code to play with.

smcantab:
[...]

PR #61, opened 2023-04-20
Edgetracking algorithm

reykboerner:
Implementation of the edgetracking algorithm, as planned in issue #7

The implementation has been tested for the 2D bistable FitzHugh-Nagumo model (test added), a 5D ocean box model, and a 5D chaotic system of a 2D ocean coupled to a 3D atmosphere.

The code features some output control but this could be expanded. Also, currently the code has only been used for systems where one is interested in the edge between two adjacent basins of attraction. If there are more than two attractors with possibly fractal basin boundaries, the code should still work (raising a warning if a third basin is involved) but it becomes less clear which edge one actually wishes to track. I have some ideas how this could be addressed, happy to discuss.


Datseris:
Great, thanks a lot! I am reviewing this! I hope to submit a review by the end of the weekend!

awage:
Hi! This is quite extraordinary. I have an implementation of my own of this technique! Originally it was called saddle-straddle by Yorke in his book on the software Dynamics. We made a paper with this stuff, look at figure 1 in this paper: 
https://arxiv.org/pdf/1901.10728.pdf 

I had no idea that it was used in other context as well.

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+* `output_level = 2`: returns `[track, bisect1, bisect2]`, where `track` is the track along  
+  the edge and `bisect1`, `bisect2` are vectors containing the bisection points on either  
+  side of the basin boundary
+
+!!! warning
+    May behave erroneously when the DiffEq solver of `ds` is set to `alg = SimpleATsit5()`,
+    which is the default solver in DynamicalSystems. The recommended solver is `Vern9()`.
+
+## References
+[^1]: [Skufca et al. (2006)](https://journals.aps.org/prl/pdf/10.1103/PhysRevLett.96.174101?casa_token=RUn26KnFdNEAAAAA%3AoXsTlmEWVMkEYbOtR-j2PH2vYOOPOy1a2R_37ncnf4gsiHp6GR66M-IBpzXocLoMQC_oHhk8MIFRa_8)
```
awage, line 55: These links don't work for me. Provide de doi links please

reykboerner:
Thanks @awage, I wasn't aware that the method actually dates that far back! I should correct the references then and add the original Yorke papers. It seems that several people have implementations of this algorithm but usually for their system of interest.

I also read in your paper that you find the boundary between one basin and all other basins. To make the present code more general, one could add keyword arguments to specify two sets of attractors (by giving their respective Dict keys), so that the algorithm tracks the boundary separating these two sets.

awage:
> > Hi! This is quite extraordinary. I have an implementation of my own of this technique!
> 
> Hah, what a small world!!!

This paper I wrote was my first serious work in Julia and the seed of Attractors.jl. So-... Cheers!

awage:
> Thanks @awage, I wasn't aware that the method actually dates that far back! I should correct the references then and add the original Yorke papers. It seems that several people have implementations of this algorithm but usually for their system of interest.
> 
> I also read in your paper that you find the boundary between one basin and all other basins. To make the present code more general, one could add keyword arguments to specify two sets of attractors (by giving their respective Dict keys), so that the algorithm tracks the boundary separating these two sets.

I will give you more detail about the original work. Also, they baked in the concept of "general basins", which is simply the combination of different basins. This is more general as you suggest, but I don't know if it is really useful unless in margin cases such as the one we used in the paper. We can discuss this.
 
I have a  very  simple trivial test for the algorithm, this is the equation of the cubic map: 

$x_{n+1} = μ x_n − x_n^3$

they have two stable fixed point and and a saddle at x=0 for \mu > 1 . The algorithm should find the saddle straight away. We like simple predictable tests in DynamicalSystems.jl

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+[^3]: [Luarini and Bodai (2017)](https://iopscience.iop.org/article/10.1088/1361-6544/aa6b11)
+"""
+function edgetracking(ds::DynamicalSystem, u1, u2, attractors::Dict;
+    eps1=1e-7,
+    eps2=1e-6,
+    maxiter=100,
+    abstol=0.0,
+    tmax=Inf,
+    dt=0.01,
+    ϵ_mapper=0.1,
```
Datseris, line 66: 0.1 is a rather arbitrary value for this distance. Why do you even need to set one? Isn't the default that `AttractorsViaProximity` creates a much better choice overall?

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+
+The algorithm is initialized from two states `u1`, `u2` in state space that must belong to
+different basins of attraction. The system's `attractors` are specified as a dictionary of
+`StateSpaceSet`s.
+
+## Keyword arguments
+* `eps1 = 1e-7`: bisection distance threshold
+* `eps2 = 1e-6`: trajectory divergence distance threshold
+* `maxiter = 100`: maximum number of iterations before the algorithm stops
+* `abstol = 0.0`: convergence threshold for returned edge state (distance in state space)
```
Datseris, line 15: I don't understand what `abstol` is from this sentence, and you don't explain it more in Description either.

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+    edgestate = (u1 + u2)/2
+    
+    if output_level>0
+        bisect_pts1 = [u1]
+        bisect_pts2 = [u2]
+        edge = [edgestate]
+    end
+    verbose && println("... Iteration 1: Edge at $(edgestate)")
+
+    correction = abstol + 1
```
Datseris, line 105: This is a weird staritng value for `correction`. Use `Inf` instead.

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+            if idx_new != idx2
+                @warn "New bisection point belongs to a third basin of attraction."
+            end
+        end    
+        distance = dist(u1-u2)
+    end
+    [u1, u2]
+end;
+
+dist(x) = sqrt(sum(abs, x.^2));
```
Datseris, line 194: Your `dist` is literally `LinearAlgebra.norm` so you most defnitely don't have to define this :D

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+    end
+    verbose && println("... Iteration 1: Edge at $(edgestate)")
+
+    correction = abstol + 1
+    counter = 1
+    # edge track iteration loop
+    while (correction > abstol) & (maxiter > counter)
+        reinit!(pds, [u1,u2])
+        state = edgestate
+        distance = dist(current_state(pds, 1) - current_state(pds, 2))
```
Datseris, line 111: Alright, this is a comment I'll only leave once here, but it is an absolutely crucial comment that you need to take into account every single time you are doing an operation between the two states. **Your operations are written in suboptimal performance because they would always allocate new vectors if the state of the system is `Vector`**. Said differently, if `x = [1, 2, 3]` then the operation `x + x` is allocating a **new vector and this is unecessary because you don't actually use this vector anywhere, you just take its norm**.

You need to make a function `diffnorm` (for difference norm) that does an explicitly loop, like e.g., we do here:

```julia
function diffnorm(ds::ParallelDynamicalSystem)
    u1 = current_state(ds, 1)
    u2 = current_state(ds, 2)
    # Compute euclidean dinstace in a loop (don't care about static or not)
    d = zero(eltype(u1))
    @inbounds for i in eachindex(u1)
        d += (u1[i] - u2[i])^2
    end
    return sqrt(d)
end
```

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+            u1 = u_new
+        else 
+            u2 = u_new
+            if idx_new != idx2
+                @warn "New bisection point belongs to a third basin of attraction."
+            end
+        end    
+        distance = dist(u1-u2)
+    end
+    [u1, u2]
```
Datseris, line 191: Why are you putting `u1, u2` in a vector? This unecessesarily allocates a `Vector` in memory. Just `return u1, u2`, there is no reason to pack them in a vector...

src/basins/edgetracking.jl
Datseris: But as I explained above, in fact, you won't be calling `LinearAlgebra.norm`, as you will define `diffnorm`.

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+        bisect_pts2 = [u2]
+        edge = [edgestate]
+    end
+    verbose && println("... Iteration 1: Edge at $(edgestate)")
+
+    correction = abstol + 1
+    counter = 1
+    # edge track iteration loop
+    while (correction > abstol) & (maxiter > counter)
+        reinit!(pds, [u1,u2])
```
Datseris, line 109: Dont do `reinit!(pds, [u1, u2])` becuase this unecessarily does more things than it has to and it allocates a vector when you write `[u1, u2]`. Instead, do
```julia
set_state!(pds, 1, u1)
set_state!(pds, 2, u2)
```

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+* `maxiter = 100`: maximum number of iterations before the algorithm stops
+* `abstol = 0.0`: convergence threshold for returned edge state (distance in state space)
+* `tmax = Inf`: maximum integration time of parallel trajectories until re-bisection 
+* `dt = 0.01`: integration time step
+* `ϵ_mapper = 0.1`: `ϵ` parameter in [`AttractorsViaProximity`](@ref)
+* `verbose = false`: if true, prints info while running
+* `output_level = 2`: controls what to return (see below)
+* `kwargs...`: additional keyword arguments to be passed to `AttractorsViaProximity`
+
+## Description
```
Datseris, line 23: I would like some more clarity on how the "tracking" works. I understand how the boundary is located via successive bijections. But this gives you an arbitrary point on the boundary. Why would the next step, that is called "tracking", lead to the edge state? Doesn't it imply some assumptions about the flow, specifically that the flow locally around the basin boundary is oriented towards the edge state? Is this assumption always true?

Datseris (changes requested):
Great, thanks a lot for the contribution @reykboerner . To be honest, this is surprisingly good, and this is a complement. All basic checkpoints, like documentation, tests, legitimate code that runs are checked. So the PR is really good quality, something that I don't typically expect from a first time contributor.

I'm leaving _s e v e r a l_ comments as part of my review in this PR. All comments are rather direct and to the point. This is a harsh style I hope you can get used to it quickly because it's by far the most efficient way to review PRs.

Feel free to address them all at once or step by step and ask questions whenever things are not clear. I think after you address the comments, we can have a second round of review.

---

There is one thing that bugs me: Let's try to be proactive here. Ideally, we would like to have code that finds the basin boundary. Is this possible to build upon this code base? Sure, one way to do that is via brute force based on the bijection. You pick all possible directions around a fixed point and then you biject until you find this boundary. THis has several problems: computationally unfeasible in high dim + it doesn't work in basins that are non-convex, such as those described in #60 . 

So, given that your code finds one point on the boundary, what algorithms are there that "continue" this point to find the full boundary?

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
+export edgetracking, bisect_to_edge
+
+"""
+    edgetracking(ds::DynamicalSystem, u1, u2, attractors::Dict; kwargs...)
```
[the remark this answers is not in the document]
Datseris: I think the initial guesses shoujldbe keywords which by default is what I said above.

reykboerner:
Thanks a lot @Datseris for the feedback and detailed reviews! I'm completely on board with this style of reviewing. I'll respond to each point in the next days.

Datseris:
bump!

Datseris:
bump @reykboerner

src/basins/edgetracking.jl
reykboerner: Yes, this is a core assumption underlying the edge tracking algorithm. If the flow on the basin boundary converges to some attractor, then this attractor will be the edge state you find. Typically this is a saddle, in which case the basin boundary (or a subset of it) is its stable set. There can be multiple edge states along the basin boundary, and then the algorithm will find one of them, depending on where the initial bisection locates the boundary. It is also possible that the entire basin boundary forms the edge state. For example, consider an unstable limit cycle separating two basins. The algorithm will find some point on this limit cycle and then track along it in circles until it stops.

src/basins/edgetracking.jl
reykboerner: Good to know! I've created a diffnorm function as suggested.

src/basins/edgetracking.jl
reykboerner: Also good to know, thanks!

src/basins/edgetracking.jl
reykboerner: Yes, I have defined `diffnorm`.

src/basins/edgetracking.jl
reykboerner: I agree that the default of `AttractorsViaProximity` creates a good overall choice for `ϵ`, but in some cases the user may want to set it explicitly. Since `ϵ` is an optional positional argument rather than a keyword argument, I did not know how it could be passed to `AttractorsViaProximity` without creating an explicit kwarg.

I now did this - is this a good solution?
```julia
function edgetracking(ds::DynamicalSystem, attractors::Dict;
    ...
    ϵ_mapper=nothing,
    kwargs...)
    
    ...

    if ϵ_mapper == nothing
        mapper = AttractorsViaProximity(ds, attractors; kwargs...)
    else
        mapper = AttractorsViaProximity(ds, attractors, ϵ_mapper; kwargs...)
    end

    ...
```

src/basins/edgetracking.jl
Datseris: Then you need to state this description in the `## Description` section of the docstring!

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+Runs the edge tracking algorithm[^1] for a multistable DynamicalSystem `ds`.
+
+The algorithm is initialized from two states `u1`, `u2` in state space that must belong to
+different basins of attraction. The system's `attractors` are specified as a dictionary of
+`StateSpaceSet`s.
+
+## Keyword arguments
+* `eps1 = 1e-7`: bisection distance threshold
+* `eps2 = 1e-6`: trajectory divergence distance threshold
```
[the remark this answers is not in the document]
reykboerner: I called them `bisect_thresh` and `diverge_thresh` now.

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+* `maxiter = 100`: maximum number of iterations before the algorithm stops
+* `abstol = 0.0`: convergence threshold for returned edge state (distance in state space)
+* `tmax = Inf`: maximum integration time of parallel trajectories until re-bisection 
+* `dt = 0.01`: integration time step
+* `ϵ_mapper = 0.1`: `ϵ` parameter in [`AttractorsViaProximity`](@ref)
+* `verbose = false`: if true, prints info while running
+* `output_level = 2`: controls what to return (see below)
```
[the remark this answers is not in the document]
reykboerner: I have now written the function to return a triple `edge, track1, track2`, where `edge = (track1 + track2)/2`. Of course one could argue that one only needs `track1` and `track2` but their mean `edge` is the actual output of the algorithm that represents a track along the basin boundary, so I think it makes sense that this is the first return argument. However, the pseudo-trajectories `track1` and `track2` are also useful, as they are strictly within one basin of attraction (unlike `edge`) and can be used to derive further properties. Thus, I would not throw them away but return them as additional return arguments.

To get the edge state, one could write a helper function `edgestate(edge::StateSpaceSet)` that takes the output of `edgetracking` as input. But all that function would do is take the `N` last points of `edge`, where `N=1` if the edge state is a saddle point and some larger integer if the edge state is an extended set (limit cycle, chaotic set). So perhaps this function does too little to justify adding it?

src/basins/edgetracking.jl
```diff
@@ -0,0 +1,194 @@
...
+parallel integration are then used as new states `u1` and `u2` for a new bisection, and 
+so on, until a stopping criterion is fulfilled.
+
+## Output 
+Output can be controlled via the `output_level` argument.
+* `output_level = 0`: returns only the final state on the edge
+* `output_level = 1`: returns the track along the edge as a vector of states
+* `output_level = 2`: returns `[track, bisect1, bisect2]`, where `track` is the track along  
+  the edge and `bisect1`, `bisect2` are vectors containing the bisection points on either  
+  side of the basin boundary
```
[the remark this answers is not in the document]
reykboerner: I hope it is now clearer what the purpose of the edge tracking algorithm is. Regarding the use of `bisect1` and `bisect2`, now renamed to `track1` and `track2`, see my comment above.

src/basins/edgetracking.jl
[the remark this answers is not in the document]
Datseris: > However, the pseudo-trajectories `track1` and `track2` are also useful, as they are strictly within one basin of attraction (unlike `edge`) and can be used to derive further properties.

Why are they useful? What are the properties they can be used to derive?

test/basins/edgetracking_tests.jl
```diff
@@ -0,0 +1,13 @@
+using Attractors
+using Test
+using OrdinaryDiffEq
+
+@testset "edgetracking (FitzHugh-Nagumo)" begin
+    fhn(u, p, t) = SVector{2}([10*(u[1] - u[1]^3 - u[2]), -3*u[2] + u[1]])
```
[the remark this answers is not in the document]
reykboerner: Where would you like to add this example?

src/basins/edgetracking.jl
reykboerner: I hope I have clarified this better now.

src/basins/edgetracking.jl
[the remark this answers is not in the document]
Datseris: > To get the edge state, one could write a helper function `edgestate(edge::StateSpaceSet)` that takes the output of `edgetracking` as input.

I don't understand what this function is supposed to do. You told me its input, but I also need to know the output :D . You say the function would take the last `N` points. Where does the function get `N` from? How does the function decides when `N` should be 1?

src/basins/edgetracking.jl
Datseris: Simply check the source code of `AttractorsViaProximity` and get the same syntax that it uses for the default `ε` value. Make this syntax the keyword, and then just pass the keyword.

reykboerner:
> > Thanks @awage, I wasn't aware that the method actually dates that far back! I should correct the references then and add the original Yorke papers. It seems that several people have implementations of this algorithm but usually for their system of interest.
> > I also read in your paper that you find the boundary between one basin and all other basins. To make the present code more general, one could add keyword arguments to specify two sets of attractors (by giving their respective Dict keys), so that the algorithm tracks the boundary separating these two sets.
> 
> I will give you more detail about the original work. Also, they baked in the concept of "general basins", which is simply the combination of different basins. This is more general as you suggest, but I don't know if it is really useful unless in margin cases such as the one we used in the paper. We can discuss this.
> 
> I have a very simple trivial test for the algorithm, this is the equation of the cubic map:
> 
> xn+1=μxn−xn3
> 
> they have two stable fixed point and and a saddle at x=0 for \mu > 1 . The algorithm should find the saddle straight away. We like simple predictable tests in DynamicalSystems.jl

reykboerner:
> I have a very simple trivial test for the algorithm, this is the equation of the cubic map:
> ...
> they have two stable fixed point and and a saddle at x=0 for \mu > 1 . The algorithm should find the saddle straight away. We like simple predictable tests in DynamicalSystems.jl

I wanted to add this test but had an issue with `AttractorsViaProximity`.

The code below produces the error `BoundsError: attempt to access 0-element Vector{Float64} at index [1]`, which has something to do with `mapper.dist` being an empty vector (line 118 of `src/mapping/attractor_mapping_proximity.jl`:

```julia
using Attractors
cubicmap(u, p, n) = SVector{1}(p[1]*u[1] - u[1]^3)
ds = DeterministicIteratedMap(cubicmap, [1.0], [2.0])
fp = [sqrt(2)]
attrs = Dict(1 => StateSpaceSet([fp]), 2 => StateSpaceSet([-fp]))
mapper = AttractorsViaProximity(ds, attrs)
label = mapper([2.0])
```

This error also comes up when trying to run `edgetracking` for this case, since `mapper` is called repeatedly inside `edgetracking`. Could you help with this?

test/basins/edgetracking_tests.jl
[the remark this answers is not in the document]
reykboerner: The chaotic example I mentioned is the coupled Stommel-Lorenz84 model studied in [Mehling et al. (2023)](https://arxiv.org/abs/2308.16251), but it is perhaps too complicated for a test? I will look for a simpler 3D chaotic model that has bistability and a chaotic saddle.

src/basins/edgetracking.jl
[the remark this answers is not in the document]
reykboerner: > > However, the pseudo-trajectories `track1` and `track2` are also useful, as they are strictly within one basin of attraction (unlike `edge`) and can be used to derive further properties.
> 
> Why are they useful? What are the properties they can be used to derive?

They are useful because they consist of segments of actual trajectories of the system, unlike their mean, `edge`. These trajectories may capture the statistical properties of the saddle more accurately than `edge`, especially in the presence of multiple timescales.

Consider a fast-slow system, e.g. fast chaotic atmosphere and slow bistable ocean. If the Lyapunov time of the atmosphere is shorter than the time between two bisection steps, then the atmospheric parts of the two parallel trajectories may evolve quite differently in between these bisection steps, and averaging them to get `edge` may distort the statistics of atmospheric variability.

In many cases one will only need `edge` but I think the code should offer the option to output `track1` and `track2` as well. We could write an extra function for returning only `edge` but I find it nicer to have only one function `edgetracking`. What do you think?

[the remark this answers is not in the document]
reykboerner: > > To get the edge state, one could write a helper function `edgestate(edge::StateSpaceSet)` that takes the output of `edgetracking` as input.
> 
> I don't understand what this function is supposed to do. You told me its input, but I also need to know the output :D . You say the function would take the last `N` points. Where does the function get `N` from? How does the function decides when `N` should be 1?

Sorry, I wasn't clear. My suggestion was to include a helper function

`edgestate(edge::StateSpaceSet, N=1) -> edge_state`,

where `edge_state` is also of type `StateSpaceSet`. Thus the user would have to specify `N`. I don't think there is a clear automatic way to determine `N`, as the transient time until the algorithm reaches the edge state is system-dependent. All the function `edgestate` would do is cut the transient initial part based on `N` provided by the user. Should this function be added for convenience or is it unnecessary?

reykboerner:
Hi @Datseris, I think I have addressed all comments in your review, please let me know what you think!

Datseris:
I'm having a look at this now.

Datseris:
Yeah this system diverges to infinity:
```
julia> trajectory(ds, 100, [2.0])[1]
1-dimensional StateSpaceSet{Float64} with 101 points
       2.0
      -4.0
      56.0
 -175504.0
       5.405813486073056e15
      -1.5797311146339633e47
       3.942298606396109e141
     -Inf
     -Inf
     -Inf
       ⋮
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
```
Which makes sense of course, as x^3 is larger than x when x > 1..............

With https://github.com/JuliaDynamics/Attractors.jl/pull/108 now `mapper([2.0])` will correctly return -1 as the trajectory diverged to infinity. `mapper([1.0])` corrextly returns 1 even with the current code.

Datseris (changes requested):
Okay @reykboerner , there are small changes to do and this is good to go:

1. We don't use the `[^Battelino1988]:` syntax anymore to cite papers. We use BiBTeX citations. Use `@cite` and put the citations in the .bib file. The citations will be collected and displayed automatically. (See e.g., any other docstring for how we do it)
2. Add the test of the discrete x^3 map. The problem was that your initial conditions were already outside the basins of attraction. Use smaller starting x values and it should work.

reykboerner:
Hi @Datseris, thanks for taking another look. I have updated the `edgetracking` docstring/citations as requested and added the cubic map test. I also streamlined the `edgetracking` code to save only `track1`, `track2` while iterating, computing `edge` from these two tracks at the very end.

Great to hear that you are happy with the rest. I have one more suggestion:

For debugging or visualization purposes, it could be useful to also save the time points of re-bisections. To do this I would suggest that `edgetracking` returns `edge, track1, track2, time, bisect_idx`, where `time` is a Vector of time points of the trajectories `edge`/`track1`/`track2` and `bisect_idx` is a Vector of time indices at which a re-bisection was done.

If these are too many return arguments, we could also have the low-level function `edgetracking(pds::ParallelDynamicalSystem, mapper::Attractormapper)` return everything whereas the high-level function `edgetracking(ds::DynamicalSystem, mapper::Attractormapper)` returns only `edge`.

reykboerner:
> they baked in the concept of "general basins", which is simply the combination of different basins

As discussed a while ago with @awage, in multistable systems with >2 attractors one might want to track the boundary between combinations of multiple basins, rather than just two basins. I think this would be possible with the current code by specifying this in the Dict `attractors`, e.g. for five attractors `a1 ... a5`:
```julia
attractors = Dict(1 => StateSpaceSet([a1, a2, a3]), 2 => StateSpaceSet([a4, a5]))
```

Thus, I think we don't need to add any additional functionality to `edgetracking` in order to generalize this.

Datseris:
> debugging or visualization purposes

We have already tested the function, what do you mean by debugging here? For visualization, you can always copy paste the code and modify it. The purpose of the function isn't visualization really.




> > they baked in the concept of "general basins", which is simply the combination of different basins
> 
> As discussed a while ago with @awage, in multistable systems with >2 attractors one might want to track the boundary between combinations of multiple basins, rather than just two basins. I think this would be possible with the current code by specifying this in the Dict `attractors`, e.g. for five attractors `a1 ... a5`:
> 
> ```julia
> attractors = Dict(1 => StateSpaceSet([a1, a2, a3]), 2 => StateSpaceSet([a4, a5]))
> ```
> 
> Thus, I think we don't need to add any additional functionality to `edgetracking` in order to generalize this.

I don't understand what you mean here. Have you actually tried this?  And why would this need >2 attractors? 2 attractors also have a basin boundary. In the case of 2 attractors (what we have tested now), why doesn't it find the boundary and it finds the edge state instead?

Generally speaking, the boundary is an extremely useful quantity to have. It would be great if we have an algorithm that finds the boundaries of basins of attraction. But it never seemed that what we have here can do something like this. At least, I dont' see how specifying more attractors would make the algorithm find basins instead of the edge state... I've commented already way back that we should be pro-active and try to generalize the algorithm: 

> There is one thing that bugs me: Let's try to be proactive here. Ideally, we would like to have code that finds the basin boundary. Is this possible to build upon this code base?

Datseris:
@reykboerner one more question: what does your implementation do for a system that has 3 coexisting attractors? does there exist an edge state between each pair of two attractors? (I don't think so. in many cases the 3 attarctors share 1 edge state. Will this algorithm work and go there?)

src/boundaries/edgetracking.jl
```diff
@@ -1,11 +1,31 @@
...
+
+## Fields
+* `edge::StateSpaceSet`: the pseudo-trajectory representing the edge track
+* `track1::StateSpaceSet`: the parallel pseudo-trajectory tracking the edge within basin 1
+* `track2::StateSpaceSet`: the parallel pseudo-trajectory tracking the edge within basin 2
+* `time::Vector`: time points of the above `StateSpaceSet`s
+* `bisect_idx::Vector`: indices of `time` at which a re-bisection was done
+"""
+struct EdgeTrackingResults
+    edge::StateSpaceSet
```
Datseris, line 15: `StateSpaceSet{D, T}` instead. `StateSpaceSet` is an abstract type. Parameterize `EdgeTrackingResults{D,T}`

src/boundaries/edgetracking.jl
```diff
@@ -111,8 +135,8 @@ function edgetracking(pds::ParallelDynamicalSystem, mapper::AttractorMapper;
     u1, u2 = bisect_to_edge(pds, mapper; bisect_thresh)
     edgestate = (u1 + u2)/2
 
-    track1 = [u1]
-    track2 = [u2]
+    track1, track2 = [u1], [u2]
+    time, bisect_idx = [0.], []
```
Datseris, line 139:
```suggestion
    time, bisect_idx = [0.0], Int[]
```

src/boundaries/edgetracking.jl
```diff
@@ -149,9 +176,13 @@ function edgetracking(pds::ParallelDynamicalSystem, mapper::AttractorMapper;
     
     track1 = StateSpaceSet(reduce(hcat, track1)')
     track2 = StateSpaceSet(reduce(hcat, track2)')
-    edge = StateSpaceSet((Matrix(track1) + Matrix(track2))/2)
 
-    return edge, track1, track2
+    return EdgeTrackingResults(
+        StateSpaceSet((Matrix(track1) + Matrix(track2))/2),
```
Datseris, line 181:
```suggestion
        StateSpaceSet((track1 .+ track2)./2),
```
Welcome to the world of broadcasting.

src/boundaries/edgetracking.jl
```diff
@@ -1,11 +1,31 @@
...
+    bisect_idx::Vector{Int}
+end
+
+"""
+    edgetracking(ds::DynamicalSystem, attractors::Dict; kwargs...)
 Track along a basin boundary in a dynamical system `ds` with two or more `attractors`
 in order to find an *edge state* or saddle.
-Return a pseudo-trajectory `edge` describing the boundary, derived from two shadowing
-trajectories `track1`, `track2` tracking along either side of the boundary.
+Return a pseudo-trajectory `edge` describing the boundary, derived from two
```
Datseris, line 26: You need to say something like "return the results in the form of [`EdgetrackingR...`] which contains the tracked edge state".

src/boundaries/edgetracking.jl
reykboerner: Ok!

Datseris:
bump @reykboerner

reykboerner:
Hi @Datseris, I made all changes we talked about and checked that the three testsets pass. However, I cannot request merging the PR since the docs fail to build. It seems like this is also the case on the origin main branch. Could you help please?

Datseris:
ok, thanks, i'll review the PR and have a look at the documentation build.

docs/src/index.md
```diff
@@ -10,9 +10,7 @@ using CairoMakie, Attractors
 
 ## Latest news
 
-- Our paper on the global stability analysis framework offered by Attractors.jl ([`continuation`](@ref)) and the novel continuation offered by [`RecurrencesFindAndMatch`](@ref) is published as a _Featured Article_ in Chaos (https://pubs.aip.org/aip/cha/article/33/7/073151/2904709/Framework-for-global-stability-analysis-of) and has been featured in the AIP publishing showcase (https://www.growkudos.com/publications/10.1063%25252F5.0159675/reader)
-- New function [`minimal_fatal_shock`](@ref)
-- New function [`match_continuation!`](@ref) which improves the matching during a continuation process where attractors disappear and reappear.
+- New functions [`edgetracking`](@ref) and [`bisect_to_edge`](@ref) added that implement the **edge tracking algorithm**, a method to find saddles, *edge states* or *Melancholia states* in dynamical systems, also when they are unstable chaotic sets.
```
Datseris, line 13:
```suggestion
- New functions [`edgetracking`](@ref) and [`bisect_to_edge`](@ref) added that implement an **edge tracking algorithm**, a method to find saddles or *edge states* in dynamical systems, also when they are unstable chaotic sets.
```

there are many algorithms, such as the saddle stradle, and this is one of them. (not "the")

(Also, let's stick to commonly accepted terminology and keep the docstring lean, it is fine to refer to this term in the rest of the docs)

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+    ϵ_mapper=nothing,
+    show_progress=true,
+    kwargs...)
+    
+    pds = ParallelDynamicalSystem(ds, [u1, u2])
+    mapper = AttractorsViaProximity(ds, attractors, ϵ_mapper; kwargs...)
+    
+    edgetracking(pds, mapper;
+        bisect_thresh, diverge_thresh, maxiter, abstol, T_transient, Δt, tmax, show_progress)
+end;
```
Datseris, line 127: Please remove `;` from all `end` in the code, it is unecessary.

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+"""
+function bisect_to_edge(pds::ParallelDynamicalSystem, mapper::AttractorMapper;
+    bisect_thresh=1e-6)
+    u1, u2 = current_states(pds)
+    idx1, idx2 = mapper(u1), mapper(u2)
+
+    if (idx1 == idx2)
+        if idx1 == -1
+            error("AttractorMapper returned label -1 (could not match the initial condition with any attractor).
+            Try changing the settings of AttractorsViaProximity or increasing bisect_thresh, diverge_thresh.")
```
Datseris, line 254:
```suggestion
            Try changing the settings of the `AttractorMapper` or increasing bisect_thresh, diverge_thresh.")
```

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+function diffnorm(u1, u2)
+    d = zero(eltype(u1))
+    @inbounds for i in eachindex(u1)
+        d += (u1[i] - u2[i])^2
+    end
+    sqrt(d)
+end;
+
+diffnorm(pds::ParallelDynamicalSystem) = diffnorm(current_state(pds, 1),
+    current_state(pds, 2))
```
Datseris, line 304:
```suggestion
function diffnorm(pds::ParallelDynamicalSystem)
    diffnorm(current_state(pds, 1), current_state(pds, 2))
end
```
if it takes more than 1 line might as well use the long syntax

docs/src/basins.md
```diff
@@ -29,6 +29,14 @@ basins_fractal_test
 uncertainty_exponent
 ```
 
+## Edge tracking and edge states
+The edge tracking algorithm allows to locate and construct so-called edge states (also referred to as *Melancholia states*) embedded in the basin boundary separating different basins of attraction. These could be saddle points, unstable periodic orbits or chaotic saddles. The general idea is that these sets can be found because they act as attractors when restricting to the basin boundary.
+
+```@docs
+edgetracking
+bisect_to_edge
```
Datseris, line 37:
```suggestion
edgetracking
EdgeTrackingResults
bisect_to_edge
```

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+end
+
+EdgeTrackingResults(nothing) = EdgeTrackingResults(StateSpaceSet([NaN]),
+    StateSpaceSet([NaN]), StateSpaceSet([NaN]), [NaN], [0], false)
+
+"""
+    edgetracking(ds::DynamicalSystem, attractors::Dict; kwargs...)
+Track along a basin boundary in a dynamical system `ds` with two or more attractors
+in order to find an *edge state*. Results are returned in the form of
+`EdgeTrackingResults`, which contains the pseudo-trajectory `edge` representing the track on
```
Datseris, line 32:
```suggestion
[`EdgeTrackingResults`](@ref), which contains the pseudo-trajectory `edge` representing the track on
```

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+the basin boundary, along with additional output (see below).
+
+The system's `attractors` are specified as a Dict of `StateSpaceSet`s, as in
+[`AttractorsViaProximity`](@ref) or the output of `extract_attractors`. By default, the
```
Datseris, line 36:
```suggestion
[`AttractorsViaProximity`](@ref) or the output of [`extract_attractors`](@ref). By default, the
```

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+    StateSpaceSet([NaN]), StateSpaceSet([NaN]), [NaN], [0], false)
+
+"""
+    edgetracking(ds::DynamicalSystem, attractors::Dict; kwargs...)
+Track along a basin boundary in a dynamical system `ds` with two or more attractors
+in order to find an *edge state*. Results are returned in the form of
+`EdgeTrackingResults`, which contains the pseudo-trajectory `edge` representing the track on
+the basin boundary, along with additional output (see below).
+
+The system's `attractors` are specified as a Dict of `StateSpaceSet`s, as in
```
Datseris, line 35:
```suggestion
The system's `attractors` are specified as a `Dict` of `StateSpaceSet`s, as in
```

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+the parallel integration are then used as new states `u1` and `u2` for a new bisection, and 
+so on, until a stopping criterion is fulfilled. 
+
+Two stopping criteria are implemented via the keyword arguments `maxiter` and `abstol`.
+Either the algorithm stops when the number of iterations reaches `maxiter`, or when the
+state space position of the updated edge point changes by less than `abstol` (in
+Euclidean distance) compared to the previous iteration. Convergence below `abstol` happens
+after sufficient iterations if the edge state is a saddle point. However, the edge state
+may also be an unstable limit cycle or a chaotic saddle. In these cases, the algorithm will
+never actually converge to a point but (after a burn-in) continue populating the set
```
Datseris, line 83: What does "burn-in" mean? I have not heard this expression before.

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+## Output
+
+Returns a data type `EdgeTrackingResults` containing five fields:
+* `edge::StateSpaceSet`: the pseudo-trajectory representing the tracked edge segment
+  (given by the average in state space between `track1` and `track2`)
+* `track1::StateSpaceSet`: the pseudo-trajectory tracking the edge within basin 1
+* `track2::StateSpaceSet`: the pseudo-trajectory tracking the edge within basin 2
+* `time::Vector`: time points of the above `StateSpaceSet`s
+* `bisect_idx::Vector`: indices of `time` at which a re-bisection occurred
+* `success::Bool`: indicates whether the edge tracking has been successful or not
```
Datseris, line 102:
```suggestion
Returns a data type [`EdgeTrackingResults`](@ref) containing the results.
```

Let's please avoid duplication of information. The type docstring already describes this.

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+
+"""
+    edgetracking(pds::ParallelDynamicalSystem, mapper::AttractorMapper; kwargs...)
+Low-level function for running the edge tracking algorithm, see [`edgetracking`](@ref)
+for a description, keyword arguments and output type.
+
+`pds` is a `ParallelDynamicalSystem` with two states. The `mapper` must be an
+`AttractorMapper` of subtype `AttractorsViaProximity` or `AttractorsViaRecurrences`.
+"""
+function edgetracking(pds::ParallelDynamicalSystem, mapper::AttractorMapper;
```
Datseris, line 137: There should be a `verbose = true` keyword that enables the warnings, so that they can be silenced by the user with `verbose = false`. A user that anyways checks the success return code will not care about thrown wornings on the console. The keuword should also be used in `bisect_to_edge`.

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+    tmax=Inf,
+    show_progress=true)
+    
+    if bisect_thresh >= diverge_thresh
+        error("diverge_thresh must be larger than bisect_thresh.")
+    end
+
+    # initial bisection
+    u1, u2, success = bisect_to_edge(pds, mapper; bisect_thresh)
+    if ~success
```
Datseris, line 153:
```suggestion
    if !success
```
`!` is more typical for boolean not.

src/boundaries/edgetracking.jl
Datseris: this keyword should also silence the success printing if false.

src/boundaries/edgetracking.jl
```diff
@@ -0,0 +1,304 @@
...
+            if T >= T_transient
+                push!(track1, current_state(pds, 1))
+                push!(track2, current_state(pds, 2))
+                push!(time, T)
+            end
+        end
+        # re-bisect
+        u1, u2, success = bisect_to_edge(pds, mapper; bisect_thresh)
+        if ~success
+            track1 = StateSpaceSet(reduce(hcat, track1)')
```
Datseris, line 184: Why are you doing `reduce(hcat, track1)'` instead of just giving in `track1`...? `StateSpaceSet` accepts a vector of vectors as an input directly, and that's what `track1` is.

Datseris (commented):
Yes !!! SO CLOSE! ITS HAPPENING

 I have some comments left. 

And, I have another comment: we also discussed that you would write a small example in the Examples in the documentation. Can you please copy one of the tests, hopefully the most visually appealing one, and make it into an example as well?

Meanwhike, I will try to fix why the test builds fail.

src/boundaries/edgetracking.jl
reykboerner: To me "burn-in" is a common term to refer to the transient period after initialization of a simulation that is thrown away if one is interested in some kind of converged solution or steady-state. I will change it to "transient period" to be more explicit, ok?

Datseris: yes

src/boundaries/edgetracking.jl
reykboerner: Okay, done. Is there a smarter way to silence warnings than
```
if verbose
    @warn(...)
end
```
?

Datseris:
I've fixed the failing docuentation build.

reykboerner:
Hi @Datseris, thanks for fixing the docs! I've added the example - any further changes needed before merging?

Datseris (approved):

Datseris:
Great, thanks a lot @reykboerner . I am merging this in. I will tag a new release along with #120 !

reykboerner:
> Great, thanks a lot @reykboerner . I am merging this in. I will tag a new release along with #120 !

Thanks @Datseris, I'm glad I could contribute and learned a lot! There still seems to be an issue with the docs - I think running the code in my example requires the missing `using OrdinaryDiffEq` because I am specifying `alg=Vern9()` in the `CoupledODEs` definition of `fitzhugh_nagumo`.

How does using additional packages work in the docs examples? Should `OrdinaryDiffEq` be added to the deps in `docs/Project.toml` and then the line `using OrdinaryDiffEq` added in my example in `docs/src/examples.md`?

If this is correct I can make these changes and make another PR to include them.

Datseris:
yes thats correct. Ordinary DIffeq should already be in `docs/Project.toml` so just the using should be missing

Datseris:
(i was going to check docs after merging all other PRs and before a release, but thanks for catching this early)

PR #66, opened 2023-04-26
Cite arxiv paper

Datseris:


awage:
> cc @awage I just tried to run the example you have coded as part of the extensive tests:

Ok, I'll check what is going on.

awage:
The system is OK, it comes from this paper: 
Captura de pantalla_2023-04-26_17-13-09

There is a problem with the clustering parameters.

awage:
It was a problem with the parameters of the mapper and the featurizer due to the changes in rescaling of the features.

Issue #73, opened 2023-06-01
Key `0` returned by `AttractrosViaRecurrences`


Datseris:
**Describe the bug**

If my understanding is correct, the output of `basins_fractions` or `basins_of_attraction` with `AttractorsViaRecurrences` **should never contain `0` as an attractor ID. Only `-1` and the positive integers**. Yet, here is an example where `0` is a key in the found basins!

**Minimal Working Example**
```julia
using Attractors

function dissipative_standard_map_rule(u, p, n)
    x, y = u
    ν, f₀ = p
    s = x + y
    xn = mod2pi(s)
    yn = (1 - ν)*y + f₀*(sin(s))
    return SVector(xn, yn)
end

p0 = (ν = 0.02, f0 = 4.0)
u0 = [0.1, 0.1]
ds = DeterministicIteratedMap(dissipative_standard_map_rule, u0, p0)
density = 2000
xg = range(0, 2π; length = density+1)[1:end-1]
yg = range(-ymax, ymax; length = density)
grid = (xg, yg)

mapper_kwargs = (
    mx_chk_att = 3,       # attractors are well-separated in the state space: so this is small
    mx_chk_fnd_att = 200, # attractors are low period orbits, so this is small
    mx_chk_loc_att = 200, # very high multistability means this should be high
    mx_chk_hit_bas = 10,  # for more accuracy, but in any case basins are so riddled
                          # that it is unlikely that this clause will even be hit...
    sparse = false, # we want to compute full basins
    Ttr = 10,
)

mapper = AttractorsViaRecurrences(ds, grid; mapper_kwargs...)

basins, attractors = basins_of_attraction(mapper)
ids = sort!(unique(basins))
```

```
142-element Vector{Int32}:
  -1
   0
   1
   2
   3
   4
   5
   6
   7
   8
   ⋮
```

**Package versions**

Latest stable.

---

This issue also highlights that a logging system is useful (discussed in #70 )

awage:
Using `sparse = true` gives the correct answer. That's a problem with the matrix in `bsn_nfo`. I'll check this

Datseris:
This means that there are some cells whose initial conditions are not used (because by defualt teh grid isinitialized with 0s everywhere when `sparse = false`). But why would this be? Why would some cells be skipped?

awage:
Shouldn't happen. The key 0 is not present in `attractors`. Some cells are not initialized or they are skipped for some reason.

PR #75, opened 2023-06-01
bugfix for basins of attraction

awage:
As mentioned in issue  #73 basins is not updated correctly. The reason is that the structure `mapper.bsn_nfo.basins` is not updated anymore with the basins value since the switch to the new interface with sparse matrix. 

The fix will work for now. But we should consider updating the structure `mapper.bsn_nfo` with the basins information. I am not sure if there is a regression issue here. We have to be careful not breaking anything.


awage:
Before merging I'd like to add a test for this. I will do it tomorrow.

src/mapping/attractor_mapping_recurrences.jl
```diff
@@ -158,7 +158,7 @@ basins, making the computation faster as the grid is processed more and more.
     [Chaos 32, 023104 (2022)](https://doi.org/10.1063/5.0076568)
 """
 function basins_of_attraction(mapper::AttractorsViaRecurrences; show_progress = true)
-    basins = mapper.bsn_nfo.basins
+    basins = zeros(Int32, size(mapper.bsn_nfo.basins))
```
Datseris, line 161: I don't see how this would fix anything; cells that weren't processed before for whatever reason, still wouldn't be processed...

awage: It does fix the problems with the zeros. The problem in the new code is that the matrix `mapper.bsn_nfo.basins` is not updated with the basins value anywhere. The function `_get_label_ic!` returns the value but does not update the structure. It keeps only attractors for matching. 

Matching works perfectly but the possible speed improvement due to the basins identification is lost. It is a easy fix but I want to check if there is no regression before patching.

awage: I have tested the code including an update of the structure `mapper.bsn_nfo_basins` with the values of the basins (as for now it only keeps the attractors in the matrix). I do not recommend to update the matrix because: 
1. It slows everything  for your example. It is 10 times slower with the update of the matrix. I think the reason is that when the finite state machine enters the `bas_hit:` state it resets the counters and the method stumbles. 
2. It can be less precise. Results are different for the example you gave. Some initial conditions are labelled differently because the trajectory may come close to other basins. When the matrix is not updated there is no shortcut and you trust the matching only when you get to the attractor. 

If we want to keep this option, the code has to be modified slightly and I suggest a keyword to switch on this option that should be off by default. If we don't find it compelling, we can remove the state `bas_hit:` from the FSM and it will work just fine.

Datseris: Okay, I see why the basin hitting clause can slow things down here since we have large multistability. I agree with you for doing a PR that adds a keyword that turns off the basin hitting clause. But we can't remove it from the source code that's a breaking change!

awage: Sorry, matching in this context is pairing an initial condition to an attractor.

Datseris: I still haven't understood though how this simple change fixes the problem #73 has with the zeros. We haven't solved the actual problem yet, we have just completely excluded some parts of the source code from running.

awage: I think the problem with  `basins = mapper.bsn_nfo.basins` is that the matrix is not deep-copied. I am not expert on object reference in Julia but it seems that the two objects are not in independent in memory space. While I was debugging I saw some entries in the matrix basins where changing values. 

While if you assign a new matrix and update it entries, it works fine. Does it makes sense?

Datseris: But that is eaxctlu the point; there is no reason to deepcopy the matrix because the Finite STate Macine takes care of putting the correct values to a given cell uisng the `get_label_ic!(` function. Right?

awage: >  the Finite STate Macine takes care of putting the correct values to a given cell uisng the `get_label_ic!(` function. Right?

Not anymore, the mapper function returns the value but does not update the internal matrix value. The reason is that you can call any initial condition you want and the value is not tied to the internal grid of the mapper. I give you an example: 

* Suppose you have defined a grid for the recurrence of 100x100 in the box [-1,1]x[-1,1]. With the old version you could have only access to these 100x100 points in this box. With the new version you can call the mapper at **any coordinates**  in the box and it will give you the corresponding attractor. It is not cached in the internal matrix. 

In a sense you loose some information of previous computed initial conditions. But the gain in flexibility and precision is baffling.

Datseris: yes sure but the update of the basins happens later in this code when we d0 `basins[ind] = get_label_ic!(`. and that's why we want to use the same matrix the mapper stores internally, so that hte internal FSM also has access to cells labelled as basins and can also use the basin convergence clause.

awage: Found the bug. Tt is a book keeping error. When the function returns -1 we have to make sure that the temporary labelled cells returns to zero. 

This fix the code for now. But I want to check if there is a better way to do this.

awage:
I tried to remove some occurences of `relabel_visited_cell!(bsn_nfo, bsn_nfo.visited_cell, 0)` but the code breaks. So it is better to leave it like that for now.

Datseris:
Sure, this seems good to me. Can we come up with a simple test for this? I guess the only scenario where this didn't work is when the initial conditions stepped too much and the algorithm decided they would not converge. How about using the map I used in #73  and then set the Maximum Amount of total iterations to something tiny, like 10, and then checking that all the basins array becomes `-1` (as would be expected)?

Datseris:
Did you run this test and it passed locally? Because it doesn't run online, it is part of the extensive suite.

awage:
Tests are passing locally. I fixed a broken test in the extensive tests suite.

Datseris:
Thanks a lot @awage

PR #78, opened 2023-06-08
Fix bug in heatmap plot

### LINKED ISSUE #77: heatmap_basins_attractors doesn't work for Henon map
The following codes:
```
using CairoMakie
using Attractors, StaticArrays
function henon_rule(u, p, n) 
    x, y = u
    a, b = p
    xn = 1.0 - a*x^2 + y
    yn = b*x
    return SVector(xn, yn)
end
u0 = [0.2, 0.3];
p0 = [1.2, 0.3];
henon = DeterministicIteratedMap(henon_rule, u0, p0);
xg = range(-2.5, 2.5; length = 1500)
yg = range(-2.5, 2.5; length = 1500)
grid=(xg,yg)
mapper = AttractorsViaRecurrences(henon, grid; sparse = false)
basins, attractors = basins_of_attraction(mapper; show_progress = false)
heatmap_basins_attractors(grid, basins, attractors)
```
give a picture like this:
plot_1
Just the attractor, no basins. I am using Julia 1.9.

awage:
There is a small bug in `heatmap_basins_attractor`. Please review, maybe it will not work for other cases. 

Closes #77

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "Attractors"
 uuid = "f3fd9213-ca85-4dba-9dfd-7fc91308fec7"
 authors = ["George Datseris <datseris.george@gmail.com>", "Kalel Rossi", "Alexandre Wagemakers"]
 repo = "https://github.com/JuliaDynamics/Attractors.jl.git"
-version = "1.6.0"
+version = "1.6.1"
 
 [deps]
 Clustering = "aaaa29a8-35af-508c-8bc3-b662a17a0fe5"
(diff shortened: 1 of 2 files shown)


Datseris (changes requested):
I don't think this is correct. The colormap is what it should be. We obtain colors using the positive integers. We create two maps: one is a map from the positive integers to the attractor keys (ie., enumerating the keys) and the other is from the positive integers to the colors.

I don't think this PR solves something, but it for sure breaks previously working examples.

awage:
I understand now: -1 should be treated separately to fix the problem.

awage:
I changed the color range for the -1 case. It should fix the problem.

Datseris:
ok this finally works and is the solution. The colormap _values_ are the positive integers. This is true for all categorical colormaps that are made by default in makie. you'd need to make a custom colormap if you want the _values_ to not bee the positive integers.

The solution is to simply transform the basin array to also use the positive integers as keys. This will work now even for keys `[-1, 32, 2, 85871779]`.

PR #80, opened 2023-06-26
Brute force mfs draft, BlackBoxOptim in progress

StefanVaylBX2023:


src/mfs.jl
```diff
@@ -0,0 +1,163 @@
...
+
+function check_basin(mapper::AttractorMapper, X, Y)
+    idx = mapper(X)
+    idy = mapper(Y)
+    if idx == idy
+        return true
+    else
+        return false
+    end
+end
```
Datseris, line 26: This function needs to be removed. First, there is no reason to dedicate a new function and 8 lines of code to what is literally the operation `mapper(X) == mapper(Y)`. But most importantly, `Y` here is always the same, it is the starting initial condition. There is no reason to calculate its mapped id more than once, so do that and save the mapped id and then just compare `mapper(X) == id_u0`.

src/mfs.jl
```diff
@@ -0,0 +1,163 @@
...
+    coordinates of a random point on the n-sphere
+
+Description:
+    This function generates a random point on the n-sphere of radius radius.
+    It does so by generating n-1 random angles and then calculating the last coordinate
+    using the formula for the n-sphere.
+"""
+
+
+function random_point_on_sphere(radius, n)
```
Datseris, line 45: There is a more efficient way to do this, that we use in StateSpaceSets.jl, in the sampler function: https://github.com/JuliaDynamics/StateSpaceSets.jl/blob/0f0e5224e3d7b2922bb6e408911ad932cbe22ef5/src/sampler.jl#L88-L103 You should utilize the `statespace_sampler` function directly so that we reduce the amount of total code written. You can re-initialize a random sampler via `statespace_sampler` each time the radius changes.

src/mfs.jl
```diff
@@ -0,0 +1,163 @@
...
+    best_shock: vector of the best shock found
+    best_dist: norm of the best shock found
+
+Description:
+    This function runs the minimal fatal shock algorithm on the initial point X.
+    It does so by first generating a random pertubation of X and then trying to find a better one by using the brute force method.
+"""
+
+
+function mfs(mapper::AttractorMapper, search_area, X, dim, n_iterations = 10000)
```
Datseris, line 158: Please rename to `minimal_fatal_shock`. Also make `n_iterations` a keyword, and name it something more intuitive, like `total_iterations`. ("n" means nothing, `n_iterations` is just as descriptive and intuitive as `iterations`).

src/mfs.jl
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    return best_shock, best_dist
+end
+
+
+"""
+    mfs(mapper, search_area, X, dim, n_iterations = 10000)
+
+Input:
```
Datseris, line 141: Please adjust all documentation strings according to DynamicalSystems.jl: https://juliadynamics.github.io/DynamicalSystems.jl/latest/contributors_guide/#Documentation-string-style

src/mfs.jl
```diff
@@ -0,0 +1,163 @@
...
+
+Description:
+    This function runs the minimal fatal shock algorithm on the initial point X.
+    It does so by first generating a random pertubation of X and then trying to find a better one by using the brute force method.
+"""
+
+
+function mfs(mapper::AttractorMapper, search_area, X, dim, n_iterations = 10000)
+    best_shock, best_dist = random_pertubation(mapper, X, search_area, dim,  n_iterations)
+    best_shock, best_dist = mfs_brute_force(mapper, X, best_shock, best_dist, dim, n_iterations)
```
Datseris, line 160: We should be proactive here. We should re-work the `minimal_fatal_shock` main function to use multiple dispatch, in anticipation of different algorithms being used to find the minimal perturbation. I propose that the fourth argument is a new Type that indicates what kind of minimization process to use. Its options would be brute force, black box optimization, or NOMAD. At the moment we have the brute force version implemented. 

When this new Type is initialized, its initialization also contains all keywords related to the specific optimization approach, such as brute force or blackbox.

By the way, there is no reason to ask for `dim` as input, as it is `dim = dimension(mapper.ds)`.

src/mfs.jl
```diff
@@ -0,0 +1,163 @@
...
+    
+    radius_dif = 100.
+
+    for _ in 1:Ntries
+
+        pertubation = random_point_on_sphere((best_dist-best_dist/radius_dif), dim)
+        new_shock = X + pertubation
+        
+        if !(check_basin(mapper, X, new_shock))
+            best_dist = set_distance(StateSpaceSet([X]), StateSpaceSet([new_shock]), Centroid())
```
Datseris, line 129: Why is `StateSpaceSet([X])` created all the time inside the loop of `Ntries`? Create it once outside the loop and re-use it.

src/mfs.jl
```diff
@@ -0,0 +1,163 @@
...
+
+Description:
+    This function works on the results obtained by random_pertubation(). It starts from the best shock found so far and tries to find a better one 
+    by continuously reducing the radius of the sphere on the surface of which it generates random pertubations. If pertubation with the same basin of attraction is found,
+    it updates the best shock found so far and reduces the radius of the sphere. 
+    It repeats this process Ntries times and returns the best pertubation found.
+"""
+
+
+function mfs_brute_force(mapper::AttractorMapper, X, best_shock, best_dist, dim,  Ntries=10000)
```
Datseris, line 119: This function should utilize a new Type created for the `minimal_fatal_shock` main function (see comment below). That type must contain all information related with the brute force algorithm , like the number of tries used in the initial random perturbation search, as well as the number of tries used in the "minimization" step of successively reducing the hypersphere radius.

src/mfs.jl
```diff
@@ -0,0 +1,163 @@
...
+    
+    radius_dif = 100.
```
Datseris, line 121: Where does the value `100` come from?

src/mfs.jl
Datseris: Actually, `n_iterations` should become an argument of the new Type that will specify the Brute Force approach, see below.

Datseris (changes requested):
Alright, this is a good start! But there are many things to improve as well!

StefanVaylBX2023:
HI George!
I am very sorry for last minute changes, but I just got an appointment to
collect some important documents, that's the only possible time slot, can
we reschedule the meeting for the other day?
I will PR all the changes you proposed in your email today +  currently I
am working on construction of optimization with blackboxoptim.jl, where i
use objective function with penalties (as there is indeed no direct
application of blackbox constraint), this thing seems to be promising,
while the approach in the paper, as we already discussed, is not applicable
since they derive equations locally using lagrange multipliers approach for
each case separately.
I am very sorry for inconvenience.

function mfs_objective(perturbation, x, idx, mapper, penalty=100.0)

dist = LinearAlgebra.norm(perturbation)
if dist == 0
return penalty
end
if mapper(x + perturbation) == idx
return dist + penalty
else
return dist
end
end




вт, 27 июн. 2023 г. в 00:16, George Datseris ***@***.***>:

src/mfs.jl
```diff
@@ -0,0 +1,170 @@
...
+X: initial point to be tested
+search_area: array of two points defining the search area
+dim: dimension of the system
+total_iterations: number of random pertubations to try, default = 10000
+
+Output:
+    best_shock: vector of the best shock found
+    best_dist: norm of the best shock found
+"""
+
```
Datseris, line 33: Why are you leaving an empty line between the docstrig and the function?

src/mfs.jl
```diff
@@ -0,0 +1,170 @@
...
+        end
+    end
+
+    return best_shock, best_dist
+end
+
+
+
+
+function mfs_objective(perturbation, X, idX, mapper::AttractorMapper, penalty=1000.0)
```
Datseris, line 98: where does 1000.0 come from and why is not not specified by the user?

src/mfs.jl
```diff
@@ -0,0 +1,170 @@
...
+    total_iterations: number of iterations to run each step of the algorithm, default = 10000
+    algorithm: random, bboxopt or combined.  
+Output:
+    best_shock: vector of the best shock found
+    best_dist: norm of the best shock found
+
+
+"""
+
+function minimal_fatal_shock(mapper::AttractorMapper, X, search_area; dimension, total_iterations = 10000, algorithm = :random)
```
Datseris, line 138: I said that we should use multiple dispatch. Create types that correspond to the algorithm possiblities and in these types the user can provide deciding arguments. Like, which algorithm from blackboxoptim to use, or how large the penantly wshould be, or how many iterations to do. Everything should be tunable by the user by tuning the fields of these new types that decide the algorithm.

Datseris (commented):
The previous PR was closed and a new one was opened but many comments have not been resolved. This also needs tests in the `test` folder.

PR #83, opened 2023-07-04
Minimal fatal shock algorithm

StefanVaylBX2023:
Hey!

Here is working code with blackbox optimization of mapper, tests for newton fractal showed more accurate results than with random algo more testing and modifications in process


src/mfs.jl
```diff
@@ -0,0 +1,170 @@
+
+
+
+function random_point_on_sphere(radius, n)
```
Datseris, line 4: There is a more efficient way to do this, that we use in StateSpaceSets.jl, in the sampler function: https://github.com/JuliaDynamics/StateSpaceSets.jl/blob/0f0e5224e3d7b2922bb6e408911ad932cbe22ef5/src/sampler.jl#L88-L103 You should utilize the statespace_sampler function directly so that we reduce the amount of total code written. You can re-initialize a random sampler via statespace_sampler each time the radius changes.

src/mfs.jl
```diff
@@ -0,0 +1,170 @@
...
+X: initial point to be tested
+search_area: array of two points defining the search area
+dim: dimension of the system
+total_iterations: number of random pertubations to try, default = 10000
+
+Output:
+    best_shock: vector of the best shock found
+    best_dist: norm of the best shock found
+"""
+
```
Datseris, line 33: Why are you leaving an empty line between the docstring and the function?

src/mfs.jl
```diff
@@ -0,0 +1,170 @@
...
+    best_shock: vector of the best shock found
+    best_dist: norm of the best shock found
+
+
+"""
+
+
+function mfs_brute_force(mapper::AttractorMapper, X, best_shock, best_dist, dim, id_x,  total_iterations=10000)
+    
+    radius_dif = 100.
```
Datseris, line 79: Where does the value 100 come from?

src/mfs.jl
```diff
@@ -0,0 +1,170 @@
...
+    
+    if length(search_area) == 1
+        search_area = [search_area[1] for _ in 1:dimension]
+    end
+
+    function objective_function(perturbation)  # we need this declaration here due to BlackBoxOpt specifics of input
+        return mfs_objective(perturbation, X, id_x, mapper)
+    end
+
+    if algorithm == :random
```
Datseris, line 151: I said that we should use multiple dispatch. Create types that correspond to the algorithm possiblities and in these types the user can provide deciding arguments. Like, which algorithm from blackboxoptim to use, or how large the penantly wshould be, or how many iterations to do. Everything should be tunable by the user by tuning the fields of these new types that decide the algorithm.

Datseris (commented):
I am not sure why you closed the PR and opened a new one, but this was confusing. In any case, the previous PR was closed and a new one was opened but many comments have not been resolved. This also needs tests in the test folder.

Datseris:
For next time:

- Add three test cases: newton's fractal, 3d Thomas cyclical continuous time system, Kuramoto oscillators from our paper "framework for global stability"
- Compare the implementations with respect to performance and stability (i.e., how much the MFS changes each time you run the algorithm). Also run the code for 3 different initial conditions per tested system
- Try to implement an optimization method using NOMAD.jl https://github.com/bbopt/NOMAD.jl 
- Address the PR comments

src/mfs.jl
awage: I use this for testing: 
```
u = randn(n)
u = u./norm(u)
```
But `statespace_sampler` is better.

StefanVaylBX2023:
Everything seems to be working, will write tests tmrw and do comparisons + some plots where possible. 
I've tried NOMAD but didn't succeed, the output values doesn't correspond to actual results even in 2d newton + seems to be quite computationally heavy in comparison with BlackBoxOptim.jl, I will look into their tests more tmrw to see if smth wrong with my constraints (code attached ). Testing ongoing, 2d newton works perfectly, 3d thomas works, but it is hard quite hard to analyze the results, at least at the moment both algos converge to same values

`
using Attractors
using NOMAD
using LinearAlgebra
function newton_map(z, p, n)
    z1 = z[1] + im*z[2]
    dz1 = newton_f(z1, p[1])/newton_df(z1, p[1])
    z1 = z1 - dz1
    return SVector(real(z1), imag(z1))
end
newton_f(x, p) = x^p - 1
newton_df(x, p)= p*x^(p-1)

ds = DiscreteDynamicalSystem(newton_map, [0.1, 0.2], [3.0])
xg = yg = range(-1.5, 1.5; length = 400)

newton = AttractorsViaRecurrences(ds, (xg, yg);
    sparse = false, mx_chk_lost = 1000
)

attractors = extract_attractors(newton)
basins, attractors = basins_of_attraction(newton)
newton_atr1 = ((collect(values(attractors)))[3].data)


function nomad_opt(X, mapper)

    function nomad_map(perturbation)
        return LinearAlgebra.norm(perturbation)

    end
    function nomad_constraint(perturbation)
        
        if mapper(perturbation) != mapper(X)
            return 0
        else
            return 1
        end
    end
    function eval_nomad(perturbation)
        bb_outputs = [nomad_map(perturbation), nomad_constraint(perturbation)]
        success = true
        count_eval = true
        return (success, count_eval, bb_outputs)
    end
    problem = NomadProblem(2, 
                            2, 
                            ["OBJ", "EB"], 
                            eval_nomad;
                            lower_bound=[-1.5, -1.5],
                            upper_bound=[1.5, 1.5])
    
    result = solve(problem, [X[1]+0.5, X[2]-0.5])

    return result
end

result = nomad_opt(newton_atr1[1], newton)
LinearAlgebra.norm(result.x_best_feas)


`

StefanVaylBX2023:
When trying to test Kuramoto problem always getting result with just one attractor in existence. Tested for different parameters _k_ and Nd. 

```

using Attractors
using DynamicalSystemsBase
using OrdinaryDiffEq: Tsit5, Vern9
using SparseArrays
using Random
using Graphs


mutable struct KuramotoParameters{M}
    N::Int
    α::Float64
    Δ::M
    ΔT::M
    P::Vector{Float64}
    K::Float64
    # Both of these are dummies
    x::Vector{Float64}
    y::Vector{Float64}
end
function KuramotoParameters(; N, K, α = 0.1, seed = 53867481290)
    rng = Random.Xoshiro(seed)
    g = random_regular_graph(N, 3; rng)
    Δ = incidence_matrix(g, oriented=true)
    P = [isodd(i) ? +1.0 : -1.0 for i = 1:N]
    x = Δ' * zeros(N)
    y = zeros(N)
    ΔT = sparse(Matrix(Δ'))
    return KuramotoParameters(N, α, Δ, ΔT, P, K, x, y)
end
Nd = 10
p = KuramotoParameters(; K = 4., N = Nd)
diffeq = (alg = Vern9(), reltol = 1e-9, maxiters = 1e8)
ds = CoupledODEs(second_order_kuramoto!, zeros(2*Nd), p; diffeq)

_complete(y) = (length(y) == Nd) ? zeros(2*Nd) : y;
_proj_state(y) = y[Nd+1:2*Nd]
psys = ProjectedDynamicalSystem(ds, _proj_state, _complete)
yg = range(-12, 12; length = 51)
grid = ntuple(x -> yg, dimension(psys))
mapper = AttractorsViaRecurrences(psys, grid; sparse = true, Δt = 0.01,
    show_progress = true, mx_chk_fnd_att = 100,
    mx_chk_safety = Int(1e7),
    force_non_adaptive = true,
    mx_chk_loc_att = 10)




using Distributions
random_seed = [rand(Uniform(-12, 12), Nd) for _ in 1:20]

res = map(x -> mapper(x), random_seed)

res
```

Datseris:
For the kuramoto: you didn't use the script I told you in full force. You correctly copied the Kuramoto dynamics but not the initialization of initial conditions. Here is a full script that finds attractors of Kuramoto for some given parameter. For K = 4 there are 100s of attractors, for K = 6 there are 10s:
```julia

using Attractors
using OrdinaryDiffEq: Tsit5, Vern9
using Random
using Graphs
using SparseArrays

mutable struct KuramotoParameters{M}
    N::Int
    α::Float64
    Δ::M
    ΔT::M
    P::Vector{Float64}
    K::Float64
    # Both of these are dummies
    x::Vector{Float64}
    y::Vector{Float64}
end
function KuramotoParameters(; N, K, α = 0.1, seed = 53867481290)
    rng = Random.Xoshiro(seed)
    g = random_regular_graph(N, 3; rng)
    Δ = incidence_matrix(g, oriented=true)
    P = [isodd(i) ? +1.0 : -1.0 for i = 1:N]
    x = Δ' * zeros(N)
    y = zeros(N)
    ΔT = sparse(Matrix(Δ'))
    return KuramotoParameters(N, α, Δ, ΔT, P, K, x, y)
end
using LinearAlgebra: mul!
function second_order_kuramoto!(du, u, p, t)
    (; N, α, K, Δ, ΔT, P, x, y) = p
    φs = view(u, 1:N)
    ωs = view(u, N+1:2N)
    dφs = view(du, 1:N)
    dωs = view(du, N+1:2N)
    dφs .= ωs
    mul!(x, ΔT, φs)
    x .= sin.(x)
    mul!(y, Δ, x)
    y .*= K
    # the full sine term is y now.
    @. dωs = P - α*ωs - y
    return nothing
end

Nd = 10
p = KuramotoParameters(; K = 4.0, N = Nd)
diffeq = (alg = Vern9(), reltol = 1e-9, maxiters = 1e8)
ds = CoupledODEs(second_order_kuramoto!, zeros(2*Nd), p; diffeq)

_complete(y) = (length(y) == Nd) ? zeros(2*Nd) : y;
_proj_state(y) = y[Nd+1:2*Nd]
psys = ProjectedDynamicalSystem(ds, _proj_state, _complete)
yg = range(-12, 12; length = 51)
grid = ntuple(x -> yg, dimension(psys))
mapper = AttractorsViaRecurrences(psys, grid; sparse = true, Δt = 0.01,
    show_progress = true, mx_chk_fnd_att = 100,
    mx_chk_safety = Int(1e7),
    force_non_adaptive = true,
    mx_chk_loc_att = 10)

sampler, = statespace_sampler(Random.MersenneTwister(1234);
    min_bounds = [-pi*ones(Nd) -pi*ones(Nd)], max_bounds = [pi*ones(Nd) pi*ones(Nd)]
)

res = [mapper(sampler()) for _ in 1:100]

@show unique(res)
```

src/mfs.jl
```diff
@@ -0,0 +1,230 @@
...
+`mapper`  one of available in Attractors.jl [`AttractorMapper`](@ref) constructed with respect to dynamical system
+`X` initial point to be tested
+`search_area` array of two points defining the search area
+`dim` dimension of the system
+`total_iterations` number of random pertubations to try, default = 10000
+
+"""
+
+
+function random_pertubation(mapper::AttractorMapper, X, search_area, dim, id_x;   total_iterations=10000)
```
Datseris, line 26: `random_pertubation` is an "incorrect" name. It doesn't state what the function does. The function doesn't generate a random perturbation. Probably a more appropriate name is `crude_initial_radius`

src/mfs.jl
```diff
@@ -0,0 +1,230 @@
...
+            end
+        end
+    end
+    
+    return best_shock, best_dist
+end
+
+
+""" 
+    mfs_brute_force(mapper, X, best_shock, best_dist, dim,  total_iterations=10000)
```
Datseris, line 48: It is still not clear to me why `dim` is necessary at all.

src/mfs.jl
```diff
@@ -0,0 +1,230 @@
...
+`mapper::AttractorMapper` one of available in Attractors.jl [`AttractorMapper`](@ref) constructed with respect to dynamical system.
+`X` vector containing coordinates of initial point to be tested.
+`search_area` array of tuples, by specifying single value e.g. [(-1.5, 1.5)] you specify this value for each of the variables, or you may specify separetely 
+search area for each of your states by, for example in 2 dimensions, [(-1.5, 1.5), (-1, 1)].
+`algorithm` one of the two algorithms available: `MFSBruteForce` or `MFSBlackBoxOptim`.
+
+## `MFSBruteForce`
+`initial_iterations` number of random pertubations to try in the first step of the algorithm, default = 10000.
+`sphere_iterations` number of steps while initializing random points on hypersphere and decreasing its radius, default = 10000.
+`dimension` dimension of the system, default = 2.
```
Datseris, line 121: Why would need a dimension, and how can the dimension possibly have a default value? Don't ask the user to provide you with information you already have. You have the dimension information from the dynamical system. Additionally, information that _cannot_ have a default value cannot be a keyword argument. Here dimension cannot have a default value. There is not way to pick a default. Thankfully, we don't have to, as we already have access to the dimension information.

Datseris: `dimension(ds::DynamicalSystem)` is the function you need.

src/mfs.jl
```diff
@@ -0,0 +1,230 @@
...
+* Default parameters initialization `algorithm = MFSBruteForce()` will set parameters to:
+    - `initial_iterations = 10000`
+    - `sphere_iterations = 10000`
+    - `dimension = 2`
+    - `sphere_decrease_step = 100.0`
+
+## `MFSBlackBoxOptim`
+`dimension` dimension of the system.
+`guess` vector of initial guesses for the optimization algorithm.
+`penalty` penalty value for the objective function.
```
Datseris, line 133: What does penalty mean here? Be precise.

src/mfs.jl
```diff
@@ -0,0 +1,230 @@
...
+* Default parameters initialization `algorithm = MFSBlackBoxOptim()` will set default parameters:
+    - `dimension = 2`
+    - `guess = []`
+    - `penalty = 1000.0`
```
Datseris, line 137: Why do you repeat the keywords now but with their defaults? Doesn't it make more concise to do both at the same time as 

- `penalty = 100.0`: penalty value for object function (explain what this actually means)

src/mfs.jl
```diff
@@ -0,0 +1,230 @@
...
+
+## Output
+`best_shock` vector of the best shock found.
+`best_dist` norm of the best shock found.
```
Datseris, line 141: The output isn't the "best" shock, it is the "minimal fatal shock".

src/mfs.jl
```diff
@@ -0,0 +1,230 @@
...
+
+## Description
+The minimal fatal shock algorithm is used to find the smallest perturbation of the initial point X that will
```
Datseris, line 144: An advice for docstrings: when you are referring to things in code, use backticks. Don't say X, say `X`. 

Throughout DynamicalSystems.jl we refer to initial conditions or starting states as `u0`, so I recommend to do a mass rename from `X` to `u0`.

src/mfs.jl
```diff
@@ -0,0 +1,230 @@
+using BlackBoxOptim: bboptimize, best_candidate
```
Datseris, line 1: You should re-organize this file following good practices in scientific code. This means: you start with the main functions, and after the main functions and types you write all the other functions that are called. I describe this in the good scientific code workshop, block 2, slide 21: https://github.com/JuliaDynamics/GoodScientificCodeWorkshop/blob/v1.0/block2_clearcode/block2_slides.pdf 

this means, that the more inner and more small detail a function is, the later it comes in the file. Makes for a better reading experience.

Project.toml
```diff
@@ -2,18 +2,24 @@ name = "Attractors"
 uuid = "f3fd9213-ca85-4dba-9dfd-7fc91308fec7"
 authors = ["George Datseris <datseris.george@gmail.com>", "Kalel Rossi", "Alexandre Wagemakers"]
 repo = "https://github.com/JuliaDynamics/Attractors.jl.git"
-version = "1.7.1"
+version = "1.6.1"
 
 [deps]
+BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
+CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
```
Datseris, line 9: All these packages like Cairo and FFMPEG have no reason to be in the dependencies. Also ordinary diff eq and predefined systems. We don't use predefined systems in the test suties, we define everything from scratch

test/mfs/mfstest.jl
```diff
@@ -0,0 +1,130 @@
...
+
+attractors = Dict(i => StateSpaceSet([dynamic_rule(ds).magnets[i]]) for i in 1:3)
+
+mapper = AttractorsViaProximity(psys, attractors)
+
+xg = yg = range(-4, 4; length = 201)
+grid = (xg, yg)
+basins, attractors = basins_of_attraction(mapper, grid; show_progress = false)
+
+attractor3 = ((collect(values(attractors)))[3].data)
```
Datseris, line 110: accessing the `.data` field of a state space set is not good practice. Use `vec(set)` instead.

src/mfs.jl
StefanVaylBX2023: `dimension` now is computed on the entry to main function `minimal_fatal_shock` but i still leave it in auxiliary  function as argument to avoid double computations

StefanVaylBX2023:
> 



> For the kuramoto: you didn't use the script I told you in full force. You correctly copied the Kuramoto dynamics but not the initialization of initial conditions. Here is a full script that finds attractors of Kuramoto for some given parameter. For K = 4 there are 100s of attractors, for K = 6 there are 10s:
> 
> ```julia
> using Attractors
> using OrdinaryDiffEq: Tsit5, Vern9
> using Random
> using Graphs
> using SparseArrays
> 
> mutable struct KuramotoParameters{M}
>     N::Int
>     α::Float64
>     Δ::M
>     ΔT::M
>     P::Vector{Float64}
>     K::Float64
>     # Both of these are dummies
>     x::Vector{Float64}
>     y::Vector{Float64}
> end
> function KuramotoParameters(; N, K, α = 0.1, seed = 53867481290)
>     rng = Random.Xoshiro(seed)
>     g = random_regular_graph(N, 3; rng)
>     Δ = incidence_matrix(g, oriented=true)
>     P = [isodd(i) ? +1.0 : -1.0 for i = 1:N]
>     x = Δ' * zeros(N)
>     y = zeros(N)
>     ΔT = sparse(Matrix(Δ'))
>     return KuramotoParameters(N, α, Δ, ΔT, P, K, x, y)
> end
> using LinearAlgebra: mul!
> function second_order_kuramoto!(du, u, p, t)
>     (; N, α, K, Δ, ΔT, P, x, y) = p
>     φs = view(u, 1:N)
>     ωs = view(u, N+1:2N)
>     dφs = view(du, 1:N)
>     dωs = view(du, N+1:2N)
>     dφs .= ωs
>     mul!(x, ΔT, φs)
>     x .= sin.(x)
>     mul!(y, Δ, x)
>     y .*= K
>     # the full sine term is y now.
>     @. dωs = P - α*ωs - y
>     return nothing
> end
> 
> Nd = 10
> p = KuramotoParameters(; K = 4.0, N = Nd)
> diffeq = (alg = Vern9(), reltol = 1e-9, maxiters = 1e8)
> ds = CoupledODEs(second_order_kuramoto!, zeros(2*Nd), p; diffeq)
> 
> _complete(y) = (length(y) == Nd) ? zeros(2*Nd) : y;
> _proj_state(y) = y[Nd+1:2*Nd]
> psys = ProjectedDynamicalSystem(ds, _proj_state, _complete)
> yg = range(-12, 12; length = 51)
> grid = ntuple(x -> yg, dimension(psys))
> mapper = AttractorsViaRecurrences(psys, grid; sparse = true, Δt = 0.01,
>     show_progress = true, mx_chk_fnd_att = 100,
>     mx_chk_safety = Int(1e7),
>     force_non_adaptive = true,
>     mx_chk_loc_att = 10)
> 
> sampler, = statespace_sampler(Random.MersenneTwister(1234);
>     min_bounds = [-pi*ones(Nd) -pi*ones(Nd)], max_bounds = [pi*ones(Nd) pi*ones(Nd)]
> )
> 
> res = [mapper(sampler()) for _ in 1:100]
> 
> @show unique(res)
> ```

Might be a stupid question, but still i dont understand some mechanics of construction of this system, why having 10 dimensional system with dimension(mapper.ds) = 10 you feed vector of length 20 into it to get correct behaviour, how then i should tell the algo (i do obtain dimension from mapper now directly as you said)  that actual dimension of the system is not 10 but 10 by 2 or I missing something.

src/mfs.jl
```diff
@@ -0,0 +1,215 @@
...
+
+## `MFSBlackBoxOptim`
+`guess` vector of initial guesses for the optimization algorithm, `default = []`.
+`MaxSteps` maximum number of steps for the optimization algorithm, default = 10000.
+`penalty` penalty value for the objective function, allows to adjust optimization algorithm to find the minimal fatal shock, `default = 1000.0`
+
+
+## Output
+`best_shock` - minimal fatal shock.
+`best_dist` norm of the minimal fatal shock.
```
Datseris, line 35: Output should be described in the first sentences immediatelly after the function call signature. It doesn't need a section.

Datseris:
> you feed vector of length 20 into it to get correct behaviour

Ah, apologies about that. This is really unfortunate. We shouldn't have done that, but now since it works just leave it as is. AS you can see, when defining 
```julia
_complete(y) = (length(y) == Nd) ? zeros(2*Nd) : y;
```
to re-construct the "full" system dynamics of 20 dimensions, we check whether the state to recomplete from is already 20 dimensional or not.  @awage has written this, but I am not sure if it is crucial to do. We could just return the zeros function always, i.e., `_complete(y) = [zeros(Nd)..., y]`

Datseris:
> so how i should proceed, just leave this case or maybe possible solution is to come back to optional argument of dimension specified by user (by default dimension(mapper.ds) will be used)

Always use `dimension(mapper.ds)`. Don't worry about the quirk of this particular example.


Could you also redefine

```julia
_complete(y) = [zeros(Nd)..., y]
sampler, = statespace_sampler(Random.MersenneTwister(1234);
    min_bounds = [-pi*ones(Nd)], max_bounds = [pi*ones(Nd)]
)
```

to get rid of the quirk alltogether.

awage:
I think it was due to a bug that has been fixed in DynamicalSystemsBase: [typo in docstring #174](https://github.com/JuliaDynamics/DynamicalSystemsBase.jl/pull/174)

Now it is not necessary anymore. It should work just fine without.  

> ```julia
> _complete(y) = (length(y) == Nd) ? zeros(2*Nd) : y;
> ```

awage:
What version of DynamicalSystemsBase are you using?

awage:
I will check this problem.

StefanVaylBX2023:
> What version of DynamicalSystemsBase are you using?

Currently 3.2.0

awage:
> > What version of DynamicalSystemsBase are you using?
> 
> Currently 3.2.0

I'm on it.

StefanVaylBX2023:
> > > What version of DynamicalSystemsBase are you using?
> > 
> > 
> > Currently 3.2.0
> 
> I'm on it.

Thank you!

Datseris:
There aren't any bugs anywhere. Why are you using `_proj_state(y) = y[Nd+1:2*Nd]` that's not what I have written down. I have written down `_complete(y) = [zeros(Nd)..., y]`.

Datseris:
@StefanVaylBX2023 which line exactly throws this error?

StefanVaylBX2023:
> @StefanVaylBX2023 which line exactly throws this error?

updated `_complete(y) = [zeros(Nd)..., y]`

awage:
Ok, I have figured out where the problem is. My original code for the system had this odd `complete_state` function as a bypass  to allow to change all the initial conditions. The `statespace_sampler` gives 2Nd numbers and the algorithm does not what to do with the extra zeros in the `complete_state` function: `_complete(y) = [zeros(Nd); y]`. The complete state produces the state for the original system from the projected state. Now it should work: 

```julia
Nd = 10
p = KuramotoParameters(; K = 4.0, N = Nd)
diffeq = (alg = Vern9(), reltol = 1e-9, maxiters = 1e8)
ds = CoupledODEs(second_order_kuramoto!, zeros(2*Nd), p; diffeq)
_complete(y) = [zeros(Nd); y]
_proj_state(y) = y[Nd+1:2*Nd]
psys = ProjectedDynamicalSystem(ds, _proj_state, _complete)
yg = range(-12, 12; length = 51)
grid = ntuple(x -> yg, dimension(psys))
mapper = AttractorsViaRecurrences(psys, grid; sparse = true, Δt = 0.01,
    show_progress = true, mx_chk_fnd_att = 100,
    mx_chk_safety = Int(1e7),
    force_non_adaptive = true,
    mx_chk_loc_att = 10)
sampler, = statespace_sampler(Random.MersenneTwister(1234);
    min_bounds = -pi*ones(Nd), max_bounds = pi*ones(Nd)
)
res = [mapper(sampler()) for _ in 1:100]
@show unique(res)
```

With low K you should find many attractors.

StefanVaylBX2023:
> ```julia
> Nd = 10
> p = KuramotoParameters(; K = 4.0, N = Nd)
> diffeq = (alg = Vern9(), reltol = 1e-9, maxiters = 1e8)
> ds = CoupledODEs(second_order_kuramoto!, zeros(2*Nd), p; diffeq)
> _complete(y) = [zeros(Nd); y]
> _proj_state(y) = y[Nd+1:2*Nd]
> psys = ProjectedDynamicalSystem(ds, _proj_state, _complete)
> yg = range(-12, 12; length = 51)
> grid = ntuple(x -> yg, dimension(psys))
> mapper = AttractorsViaRecurrences(psys, grid; sparse = true, Δt = 0.01,
>     show_progress = true, mx_chk_fnd_att = 100,
>     mx_chk_safety = Int(1e7),
>     force_non_adaptive = true,
>     mx_chk_loc_att = 10)
> sampler, = statespace_sampler(Random.MersenneTwister(1234);
>     min_bounds = -pi*ones(Nd), max_bounds = pi*ones(Nd)
> )
> res = [mapper(sampler()) for _ in 1:100]
> @show unique(res)
> ```

i am getting same mistake for some reason, maybe it is on my side `The returned vector of complete_state must equal dimension(ds)`, which version of `dynamicalsystembase` are you using? maybe i need to update it

Datseris:
> maybe i need to update it

Go to its github cite and see the latest stable version:  https://github.com/JuliaDynamics/DynamicalSystemsBase.jl this tells you if you need to update anything.

Datseris:
it was confusing that we were not using the official API for `ProjectedDynamicalSystem`. Anyways, here is a full code that works 100% and only uses the oficial api and also makes perfect sense because states are sampled in the grid we search for attractors:

```julia

using Attractors
using OrdinaryDiffEq: Tsit5, Vern9
using Random
using Graphs
using SparseArrays

mutable struct KuramotoParameters{M}
    N::Int
    α::Float64
    Δ::M
    ΔT::M
    P::Vector{Float64}
    K::Float64
    # Both of these are dummies
    x::Vector{Float64}
    y::Vector{Float64}
end
function KuramotoParameters(; N, K, α = 0.1, seed = 53867481290)
    rng = Random.Xoshiro(seed)
    g = random_regular_graph(N, 3; rng)
    Δ = incidence_matrix(g, oriented=true)
    P = [isodd(i) ? +1.0 : -1.0 for i = 1:N]
    x = Δ' * zeros(N)
    y = zeros(N)
    ΔT = sparse(Matrix(Δ'))
    return KuramotoParameters(N, α, Δ, ΔT, P, K, x, y)
end
using LinearAlgebra: mul!
function second_order_kuramoto!(du, u, p, t)
    (; N, α, K, Δ, ΔT, P, x, y) = p
    φs = view(u, 1:N)
    ωs = view(u, N+1:2N)
    dφs = view(du, 1:N)
    dωs = view(du, N+1:2N)
    dφs .= ωs
    mul!(x, ΔT, φs)
    x .= sin.(x)
    mul!(y, Δ, x)
    y .*= K
    # the full sine term is y now.
    @. dωs = P - α*ωs - y
    return nothing
end

Nd = 10
p = KuramotoParameters(; K = 6.0, N = Nd)
diffeq = (alg = Vern9(), reltol = 1e-9, maxiters = 1e8)
ds = CoupledODEs(second_order_kuramoto!, zeros(2*Nd), p; diffeq)

complete = 1e-3randn(Nd)
proj_state = Nd+1:2*Nd

psys = ProjectedDynamicalSystem(ds, proj_state, complete)

yg = range(-12, 12; length = 51)

grid = ntuple(x -> yg, dimension(psys))

mapper = AttractorsViaRecurrences(psys, grid; sparse = true, Δt = 0.01,
    show_progress = true, mx_chk_fnd_att = 100,
    mx_chk_safety = Int(1e7),
    force_non_adaptive = true,
    mx_chk_loc_att = 10
)

# NEW SAMPLER that uses the GRID:
sampler, = statespace_sampler(grid, Random.MersenneTwister(1234))
 
res = [mapper(sampler()) for _ in 1:100]

@show unique(res)
```

Datseris:
p.s.: @StefanVaylBX2023 Obviously, don't use `complete = 1e-3randn(Nd)` in the test suite. We need reproducible outcomes. Create an RNG and write down an explicit sequence of numbers for the starting phases.

awage:
I want to figure out what is going with the `complet_state` business. I will fix it tomorrow. 

> it was confusing that we were not using the official API for `ProjectedDynamicalSystem`. Anyways, here is a full code that works 100% and only uses the oficial api and also makes perfect sense because states are sampled in the grid we search for attractors:

StefanVaylBX2023:
> it was confusing that we were not using the official API for `ProjectedDynamicalSystem`. Anyways, here is a full code that works 100% and only uses the oficial api and also makes perfect sense because states are sampled in the grid we search for attractors:
> 
> ```julia
> using Attractors
> using OrdinaryDiffEq: Tsit5, Vern9
> using Random
> using Graphs
> using SparseArrays
> 
> mutable struct KuramotoParameters{M}
>     N::Int
>     α::Float64
>     Δ::M
>     ΔT::M
>     P::Vector{Float64}
>     K::Float64
>     # Both of these are dummies
>     x::Vector{Float64}
>     y::Vector{Float64}
> end
> function KuramotoParameters(; N, K, α = 0.1, seed = 53867481290)
>     rng = Random.Xoshiro(seed)
>     g = random_regular_graph(N, 3; rng)
>     Δ = incidence_matrix(g, oriented=true)
>     P = [isodd(i) ? +1.0 : -1.0 for i = 1:N]
>     x = Δ' * zeros(N)
>     y = zeros(N)
>     ΔT = sparse(Matrix(Δ'))
>     return KuramotoParameters(N, α, Δ, ΔT, P, K, x, y)
> end
> using LinearAlgebra: mul!
> function second_order_kuramoto!(du, u, p, t)
>     (; N, α, K, Δ, ΔT, P, x, y) = p
>     φs = view(u, 1:N)
>     ωs = view(u, N+1:2N)
>     dφs = view(du, 1:N)
>     dωs = view(du, N+1:2N)
>     dφs .= ωs
>     mul!(x, ΔT, φs)
>     x .= sin.(x)
>     mul!(y, Δ, x)
>     y .*= K
>     # the full sine term is y now.
>     @. dωs = P - α*ωs - y
>     return nothing
> end
> 
> Nd = 10
> p = KuramotoParameters(; K = 6.0, N = Nd)
> diffeq = (alg = Vern9(), reltol = 1e-9, maxiters = 1e8)
> ds = CoupledODEs(second_order_kuramoto!, zeros(2*Nd), p; diffeq)
> 
> complete = 1e-3randn(Nd)
> proj_state = Nd+1:2*Nd
> 
> psys = ProjectedDynamicalSystem(ds, proj_state, complete)
> 
> yg = range(-12, 12; length = 51)
> 
> grid = ntuple(x -> yg, dimension(psys))
> 
> mapper = AttractorsViaRecurrences(psys, grid; sparse = true, Δt = 0.01,
>     show_progress = true, mx_chk_fnd_att = 100,
>     mx_chk_safety = Int(1e7),
>     force_non_adaptive = true,
>     mx_chk_loc_att = 10
> )
> 
> # NEW SAMPLER that uses the GRID:
> sampler, = statespace_sampler(grid, Random.MersenneTwister(1234))
>  
> res = [mapper(sampler()) for _ in 1:100]
> 
> @show unique(res)
> ```

works for me, thank you

Datseris:
@awage : 

> I want to figure out what is going with the complet_state business. I will fix it tomorrow.

nothing is wrong with the complete state. The `ProjectedDynamicalSystem` is well tested. We were simply _using it wrong_. If you read the current docstring of `PredefinedDynamicalSystem` you will see that whatever we were doing before was not in accordance with the API.

awage:
Sure, I used a dirty trick to bypass the initial sanity check of the API. Nevertheless I found a small bug in the API check. 

> @awage :
> 
> > I want to figure out what is going with the complet_state business. I will fix it tomorrow.
> 
> nothing is wrong with the complete state. The `ProjectedDynamicalSystem` is well tested. We were simply _using it wrong_. If you read the current docstring of `PredefinedDynamicalSystem` you will see that whatever we were doing before was not in accordance with the API.

StefanVaylBX2023:
I have some bugs while trying to build documentation manually, can you have a look at markdown files?
Rest was fixed.

test/mfs/mfstest.jl
```diff
@@ -133,5 +134,24 @@ end
 
 
 
+@testset "3D symmetry" begin
+
+    ds = Systems.thomas_cyclical(b = 0.1665)
```
Datseris, line 139: We are not using predefined dynamical systems in the tests. It is not good for long term maintainance. You should explicitly construct the dynamical system.

Project.toml
```diff
@@ -2,9 +2,10 @@ name = "Attractors"
 uuid = "f3fd9213-ca85-4dba-9dfd-7fc91308fec7"
 authors = ["George Datseris <datseris.george@gmail.com>", "Kalel Rossi", "Alexandre Wagemakers"]
 repo = "https://github.com/JuliaDynamics/Attractors.jl.git"
-version = "1.7.1"
+version = "1.6.1"
```
Datseris, line 5: We need version 1.8 and a new changelog entry in CHANGELOG.md.

docs/src/examples.md
```diff
@@ -55,6 +55,27 @@ in this case, to also get the attractors we simply extract them from the underly
...
 ```
+## Minimal Fatal Shock
+Finding Minimal Fatal Shock for some point `u0` on example of Newton's fractal attractors
+```@example MAIN
+attractors = extract_attractors(mapper)
+shocks = Dict()
+algo_bb = Attractors.MFSBlackBoxOptim(MaxSteps = 20000)
+for atr in values(attractors)
+    shocks[vec(atr)[1]] = minimal_fatal_shock(mapper, vec(atr)[1], (-1.5,1.5), algo_bb)
+
```
Datseris, line 66:
```suggestion
```

docs/src/examples.md
```diff
@@ -55,6 +55,27 @@ in this case, to also get the attractors we simply extract them from the underly
...
 attractors = extract_attractors(mapper)
 ```
+## Minimal Fatal Shock
+Finding Minimal Fatal Shock for some point `u0` on example of Newton's fractal attractors
+```@example MAIN
+attractors = extract_attractors(mapper)
+shocks = Dict()
+algo_bb = Attractors.MFSBlackBoxOptim(MaxSteps = 20000)
+for atr in values(attractors)
+    shocks[vec(atr)[1]] = minimal_fatal_shock(mapper, vec(atr)[1], (-1.5,1.5), algo_bb)
```
Datseris, line 65: define `u0 = atr[1]` and use `u0` in this line.

src/mfs.jl
```diff
@@ -0,0 +1,239 @@
...
+    MaxSteps::Int64
+    penalty::Float64
+    random_algo::MFSBruteForce
+end
+
+function MFSBlackBoxOptim(; guess = [], MaxSteps = 1000,  penalty = 1000.0, random_algo = MFSBruteForce(0,0,0))
+    MFSBlackBoxOptim(guess, MaxSteps, penalty, random_algo)
+end
+"""
+    minimal_fatal_shock(mapper::AttractorMapper, u0, search_area, algorithm::Union{MFSBruteForce, MFSBlackBoxOptim})
```
Datseris, line 55:
```suggestion
    minimal_fatal_shock(mapper::AttractorMapper, u0, search_area, algorithm)  
```

src/mfs.jl
```diff
@@ -0,0 +1,239 @@
...
+    random_algo::MFSBruteForce
+end
+
+function MFSBlackBoxOptim(; guess = [], MaxSteps = 1000,  penalty = 1000.0, random_algo = MFSBruteForce(0,0,0))
+    MFSBlackBoxOptim(guess, MaxSteps, penalty, random_algo)
+end
+"""
+    minimal_fatal_shock(mapper::AttractorMapper, u0, search_area, algorithm::Union{MFSBruteForce, MFSBlackBoxOptim})  
+
+Runs the minimal fatal shock algorithm on the initial point `u0` and outputs tuple `minimal fatal shock` and its `norm`. Two algorithms are available: [`MFSBruteForce`](@ref) and [`MFSBlackBoxOptim`](@ref).
```
Datseris, line 57: Runs -> Run, outputs -> output. This is the recommended way to write docstirngs.

src/mfs.jl
```diff
@@ -0,0 +1,239 @@
...
+[`MFSBruteForce`](@ref) is an algorithm based on randomised search with two consequent steps: random initialization and sphere radius reduction.
+[`MFSBlackBoxOptim`](@ref) is an algorithm based derivative free optimization. It uses BlackBoxOptim package to find the best shock.
```
Datseris, line 59:
```suggestion
```

You explain the algorithms in their respective docstrings so no reason to repeat here.

src/mfs.jl
```diff
@@ -0,0 +1,239 @@
...
+    MFSBruteForce(initial_iterations, sphere_iterations, sphere_decrease_step)
+end
+
+
+
+"""
+    MFSBlackBoxOptim(; kwargs...)
+
+Initialize the optimization algorithm used in `minimal_fatal_shock` function. It uses BlackBoxOptim.jl package to find the best shock.
+It is based on derivative free optimization and uses the objective function with penalties to find the minimal fatal shock.
```
Datseris, line 35: This is not enough and needs more details.

src/mfs.jl
```diff
@@ -0,0 +1,239 @@
+using BlackBoxOptim: bboptimize, best_candidate
+using Random: GLOBAL_RNG
+using StateSpaceSets: statespace_sampler
+export minimal_fatal_shock, MFSBruteForce, MFSBlackBoxOptim
+
+
+"""
+    MFSBruteForce(; kwargs...)
+
+Initialize the optimization algorithm used in `minimal_fatal_shock` function. It is based on randomised search with two consequent steps: random initialization and sphere radius reduction.
```
Datseris, line 10: This is not enough and needs more details on how the algorithm works.

src/mfs.jl
```diff
@@ -0,0 +1,239 @@
...
+
+Initialize the optimization algorithm used in `minimal_fatal_shock` function. It uses BlackBoxOptim.jl package to find the best shock.
+It is based on derivative free optimization and uses the objective function with penalties to find the minimal fatal shock.
+
+## Keyword arguments
+`guess` vector of initial guesses for the optimization algorithm, `default = []`.
+`MaxSteps` maximum number of steps for the optimization algorithm, default = 10000.
+`penalty` penalty value for the objective function, allows to adjust optimization algorithm to find the minimal fatal shock, `default = 1000.0`
+`random_algo` algorithm used to find the initial guess for the optimization algorithm, by default it is initialized as `MFSBruteForce(0,0,0)` and not used. To activate it,
+you need to initialize it with the parameters you want to use, e.g. `MFSBruteForce(1000,1000,100.0)` or `MFSBruteForce()` with default parameters.
```
Datseris, line 42: I'm confused about the default here. Use `nothing` as the default if you don't want anything to be used. 

Additionally, just use the actual keywords that are necessary from the brute force approach. there is no reason to copy all arguments of the bruteforce for the initial search here, because they are not used.

src/mfs.jl
```diff
@@ -0,0 +1,239 @@
...
+The minimal fatal shock algorithm is used to find the smallest perturbation of the initial point `u0` that will 
+lead to a different basin of attraction. It is inspired by the paper "Minimal fatal shocks in multistable complex networks"[^MFS2020] by L. Halekotte and U. Feudel.
+
+The randomized algorithm consists of two steps: random initialization and sphere radius reduction. On the first step, the algorithm generates random pertubations
+of the initial point `u0` and with the best obtained pertubation it proceeds to the second step. On the second step, the algorithm generates random pertubations 
+on the surface of the hypersphere with radius equal to the norm of the best pertubation found in the first step. It reduces the radius of the hypersphere to continue
+searching for the better result within smaller radius.  
+
+The optimization algorithm uses BlackBoxOptim.jl package to find the best shock. It is based on derivative free optimization and uses the objective function with penalties
+to find the minimal fatal shock.
```
Datseris, line 85: You should port these into their respective documentation strings for each algorithm.

src/mfs.jl
```diff
@@ -0,0 +1,239 @@
...
+
+[^MFS2020]:
+    Halekotte, L., Feudel, U. Minimal fatal shocks in multistable complex networks. 
+    [Sci Rep 10, 11783 (2020).](https://doi.org/10.1038/s41598-020-68805-6)
+[^BlackBoxOptim]:
+    BlackBoxOptim.jl Julia package by Robert Feldt [global optimization package for Julia](https://github.com/robertfeldt/BlackBoxOptim.jl.git)
```
Datseris, line 91: Don't refer to BlackBoxOptim like this, it isn't a citation. Writing simply BlackBoxOptim.jl is enough.

Datseris (changes requested):
Looks good. A small amount of simple changes are necessary. I have left comments. Besides the comments, you also have to:

- Ensure all source code lines are at most 92 characters of length.
- Ensure tests pass when you do `] test Attractors`.

src/mfs.jl
StefanVaylBX2023: What exactly do you expect here, in this version (comment with `nothing` admitted), if specified we do both random and BB algo and random gives guess for BB, it is just matter of setup if user wants this. You want to just perform initial random `shooting` without sphere decrease thing?

StefanVaylBX2023:
Screenshot 2023-07-19 at 13 28 06

Some tests outside my section are failing, rest is fixed as far as I see

Datseris:
@StefanVaylBX2023 we have a problem here. Endless walls of text are printed in the console when we run `minimal_fatal_shock`:
```
ction evals per second = 119070.53
Improvements/step = 0.18750
Total function evaluations = 10121


Best candidate found: [0.0244615, -0.101376]

Fitness: 0.104285411

Starting optimization with optimizer BlackBoxOptim.DiffEvoOpt{BlackBoxOptim.FitPopulation{Float64}, BlackBoxOptim.RadiusLimitedSelector, BlackBoxOptim.AdaptiveDiffEvoRandBin{3}, BlackBoxOptim.RandomBound{BlackBoxOptim.ContinuousRectSearchSpace}}
0.00 secs, 0 evals, 0 steps

Optimization stopped after 10001 steps and 0.08 seconds
Termination reason: Max number of steps (10000) reached
Steps per second = 129883.28
Function evals per second = 131506.66
Improvements/step = 0.17980
Total function evaluations = 10126


Best candidate found: [0.0153618, 0.0394085]

Fitness: 0.042296788

Starting optimization with optimizer BlackBoxOptim.DiffEvoOpt{BlackBoxOptim.FitPopulation{Float64}, BlackBoxOptim.RadiusLimitedSelector, BlackBoxOptim.AdaptiveDiffEvoRandBin{3}, BlackBoxOptim.RandomBound{BlackBoxOptim.ContinuousRectSearchSpace}}
0.00 secs, 0 evals, 0 steps

Optimization stopped after 10001 steps and 0.07 seconds
Termination reason: Max number of steps (10000) reached
Steps per second = 138903.19
Function evals per second = 140458.75
Improvements/step = 0.18940
Total function evaluations = 10113


Best candidate found: [-0.00562846, -0.0293217]

Fitness: 0.029857033
Starting optimization with optimizer BlackBoxOptim.DiffEvoOpt{BlackBoxOptim.FitPopulation{Float64}, BlackBoxOptim.RadiusLimitedSelector, BlackBoxOptim.AdaptiveDiffEvoRandBin{3}, BlackBoxOptim.RandomBound{BlackBoxOptim.ContinuousRectSearchSpace}}
0.00 secs, 0 evals, 0 steps

Optimization stopped after 10001 steps and 0.06 seconds
Termination reason: Max number of steps (10000) reached
Steps per second = 158746.12
Function evals per second = 160365.17
Improvements/step = 0.16620
Total function evaluations = 10103


Best candidate found: [-0.000965451, 0.000171923]

Fitness: 0.000980639

[ Info: Minimum distance between attractors computed: 1.7320508075688772
Starting optimization with optimizer BlackBoxOptim.DiffEvoOpt{BlackBoxOptim.FitPopulation{Float64}, BlackBoxOptim.RadiusLimitedSelector, BlackBoxOptim.AdaptiveDiffEvoRandBin{3}, BlackBoxOptim.RandomBound{BlackBoxOptim.ContinuousRectSearchSpace}}
0.00 secs, 0 evals, 0 steps
0.50 secs, 927 evals, 793 steps, improv/step: 0.231 (last = 0.2308), fitness=0.402771697
1.00 secs, 1861 evals, 1727 steps, improv/step: 0.214 (last = 0.2002), fitness=0.393395492
1.50 secs, 2801 evals, 2667 steps, improv/step: 0.198 (last = 0.1681), fitness=0.392434347
2.00 secs, 3754 evals, 3622 steps, improv/step: 0.186 (last = 0.1518), fitness=0.392377667
2.50 secs, 4718 evals, 4590 steps, improv/step: 0.169 (last = 0.1085), fitness=0.392374248
3.01 secs, 5672 evals, 5544 steps, improv/step: 0.160 (last = 0.1164), fitness=0.392373690
3.51 secs, 6641 evals, 6514 steps, improv/step: 0.153 (last = 0.1113), fitness=0.392373612
4.01 secs, 7620 evals, 7493 steps, improv/step: 0.151 (last = 0.1338), fitness=0.392373558
4.51 secs, 8592 evals, 8465 steps, improv/step: 0.149 (last = 0.1368), fitness=0.392373555
5.01 secs, 9586 evals, 9460 steps, improv/step: 0.150 (last = 0.1548), fitness=0.392373555

Optimization stopped after 10001 steps and 5.29 seconds
Termination reason: Max number of steps (10000) reached
Steps per second = 1891.62
Function evals per second = 1915.45
Improvements/step = 0.14910
Total function evaluations = 10127


Best candidate found: [-0.188288, -0.344245]

Fitness: 0.392373555

Starting optimization with optimizer BlackBoxOptim.DiffEvoOpt{BlackBoxOptim.FitPopulation{Float64}, BlackBoxOptim.RadiusLimitedSelector, BlackBoxOptim.AdaptiveDiffEvoRandBin{3}, BlackBoxOptim.RandomBound{BlackBoxOptim.ContinuousRectSearchSpace}}
0.00 secs, 0 evals, 0 steps
0.50 secs, 934 evals, 797 steps, improv/step: 0.246 (last = 0.2459), fitness=0.446428510
1.00 secs, 1870 evals, 1733 steps, improv/step: 0.230 (last = 0.2158), fitness=0.393107259
1.50 secs, 2812 evals, 2675 steps, improv/step: 0.223 (last = 0.2102), fitness=0.39240290
```
you have to fix this. Nothing like that should be printed by default. Maybe if the user explicitly asks for it via a `verbose` keywrod argument to `MFSBlackBox`

Datseris:
@StefanVaylBX2023 you haven't addressed my comment above. There are miriads of stuff printed on the REPL when one uses your `minimal_fatal_shock`. Please fix it ASAP so that nothing is printed _by default_. You may add an additional field/option to `MFSBlakBox` that when true it prints all this info.

https://github.com/JuliaDynamics/Attractors.jl/actions/runs/5689569164/job/15421320501?pr=83#step:8:463

Please fix this ASAP so that I can tag a release with your method!

StefanVaylBX2023:
> @StefanVaylBX2023 you haven't addressed my comment above. There are miriads of stuff printed on the REPL when one uses your `minimal_fatal_shock`. Please fix it ASAP so that nothing is printed _by default_. You may add an additional field/option to `MFSBlakBox` that when true it prints all this info.
> 
> https://github.com/JuliaDynamics/Attractors.jl/actions/runs/5689569164/job/15421320501?pr=83#step:8:463
> 
> Please fix this ASAP so that I can tag a release with your method!

Sorry, for some reason missed this comment, it is fixed now

Datseris:
@StefanVaylBX2023 I made some updates to your code. I believe you should study these updates in detail: I've improved the documentation and make the source code simpler. I also explained the keyword arguments more. I changed the sphere reduction to be a multiplicative factor that makes more sense.

The tests currently do not pass but I do not know why. The tests are not well written, and you should work on them. When a test fails, the failure should be informative: I should understand what failed and why.

The magnetic pendulum tests fail and I see as the failure message:

```
Magnetic 2D: Test Failed at c:\Users\gd419\.julia\dev\Attractors\test\mfs\mfstest.jl:144
  Expression: map((x->begin
                x[2] <= 0.395 && x[2] > 0.39
            end), values(blackbox_r)) |> all
```
this tells me nothing. It means that the expression after `@test` is to complicated to be informative. You should have only simple expressions after `@test`, like only `@test x[2] < 0.3`. Write an explicit `for` loop and call `@test` inside it.

(I haven't understood why tests fail, please tell me once you have fixed them)

Datseris:
Ah also a comment on the usage of `@testset`. This doesn't make any sense: `@testset begin ...` You should only use a testset if you give it a name, like `@testset "some name" begin`. Otherwise wrapping a code block in a testset doesn't serve any purpose.

StefanVaylBX2023:
> @StefanVaylBX2023 I made some updates to your code. I believe you should study these updates in detail: I've improved the documentation and make the source code simpler. I also explained the keyword arguments more. I changed the sphere reduction to be a multiplicative factor that makes more sense.
> 
> The tests currently do not pass but I do not know why. The tests are not well written, and you should work on them. When a test fails, the failure should be informative: I should understand what failed and why.
> 
> The magnetic pendulum tests fail and I see as the failure message:
> 
> ```
> Magnetic 2D: Test Failed at c:\Users\gd419\.julia\dev\Attractors\test\mfs\mfstest.jl:144
>   Expression: map((x->begin
>                 x[2] <= 0.395 && x[2] > 0.39
>             end), values(blackbox_r)) |> all
> ```
> 
> this tells me nothing. It means that the expression after `@test` is to complicated to be informative. You should have only simple expressions after `@test`, like only `@test x[2] < 0.3`. Write an explicit `for` loop and call `@test` inside it.
> 
> (I haven't understood why tests fail, please tell me once you have fixed them)

All the tests were passing with my final commits, let me get into this

Datseris:
yes I know tests were passsing, but I made a lot of changes :P I just am not sure what I broke with these changes!

StefanVaylBX2023:
> yes I know tests were passsing, but I made a lot of changes :P I just am not sure what I broke with these changes!

Two main issues:
1. Previously we agreed on type of output, which is tuple (vector, norm(vector)), now since you deleted norm thing, every test failed as most of the time norm was used as test factor and second argument of output, do you want to keep it the way we have norm in output (I thought it is a good feature) or should i rewrite all the tests.
2. All randomized algos were failing as the thing with a generator still mysterious for me, state_space_sampler would never return anything except vector with norm = 1.0, so the line `generator() * temp_dist` inside `mfs_brute_force` was done on purpose, without this line, if you test it, output will always have a norm = 1.0

I've studied most of your changes which indeed benefit overall structure and readability. Thank you for covering those parts which required changes.

Datseris:
> Previously we agreed on type of output, which is tuple (vector, norm(vector)), now since you deleted norm thing, every test failed as most of the time norm was used as test factor and second argument of output, do you want to keep it the way we have norm in output (I thought it is a good feature) or should i rewrite all the tests.

There is no reason to return the norm, as you can just do `norm(vector)` after the fact to obtain it. The simpler the return signature, the better.

I'll have a look at the state space sampler thing then.

Datseris:
Okay, there was a bug in the sphere state space sampler which I've fixed here: https://github.com/JuliaDynamics/StateSpaceSets.jl/pull/19

@StefanVaylBX2023 please update your StateSpaceSets.jl to version v1.3.7 and then fix the tests: since we now only return the shock vector, if you need its norm calculate it explicitly in the test suite.

Datseris:
@StefanVaylBX2023 forget about this PR for a moment. I realized that you want to sample vectors only on the surface of the hypersphere, not inside it. While the code in The StateSpaceSets.jl samples also on the inside of the sphere. I'll adjust accordingly here once I do some more adjustments in StateSpaceSets.jl.

Datseris:
@StefanVaylBX2023 Okay, I've fixed StateSpaceSets.jl. Now, the version that you should be using is 1.4.1. With this version the magnetic pendulum and 3D symmetry tests pass. However, the newton tests do not pass.

And, more importantly, you need to start adding some documenting comments to the tests. Where is the number `0.5` coming from in the test suite? For the randomized initial points, you can test that all MFS must be less than the distance between the two fixed points of the newton map.

StefanVaylBX2023:
> @StefanVaylBX2023 forget about this PR for a moment. I realized that you want to sample vectors only on the surface of the hypersphere, not inside it. While the code in The StateSpaceSets.jl samples also on the inside of the sphere. I'll adjust accordingly here once I do some more adjustments in StateSpaceSets.jl.

Yes indeed, I wanted to tell you about that tmrw because I you already fixed that, and I thought it is maybe again a way you want to this, but indeed, new sampler generates points within radius, not explicitly on the surface, which affects resultsm 

> @StefanVaylBX2023 Okay, I've fixed StateSpaceSets.jl. Now, the version that you should be using is 1.4.1. With this version the magnetic pendulum and 3D symmetry tests pass. However, the newton tests do not pass and you need to fix this.
> 
> And, more importantly, you need to start adding some documenting comments to the tests. Where is the number `0.5` coming from in the test suite? For the randomized initial points, you can test that all MFS must be less than the distance between the two fixed points of the newton map.

I just wanted to write you regarding statespace_sampler, I will check rn what's wrong with newton and add some documentation, and what exactly do you mean by `distance between two fixed points` like what should be the choice of these points?

Datseris:
fixed points = attractors of the newton map = the points the trajectories converge to. If you think about it, a MFS cannot have norm more than half the distance between any two fixed points

StefanVaylBX2023:
> fixed points = attractors of the newton map = the points the trajectories converge to. If you think about it, a MFS cannot have norm more than half the distance between any two fixed points



> fixed points = attractors of the newton map = the points the trajectories converge to. If you think about it, a MFS cannot have norm more than half the distance between any two fixed points

yeah, but 0.5 is smaller then half the minimal distance between attractors which is (~1,72/2 =0.86), and initial test was to see that if we locate a point inside this box with vertices in (0,0) and (+- 0.5, +-0.5) it cannot possibly return anything with magnitude more than 0.5, if you want i can do less random test specifying the box without bounding it to (0,0) and estimate more exact values for the test

Datseris:
okay now i understand the 0.5. You can keep the 0.5 as long as you fixed the newton tests.

StefanVaylBX2023:
> okay now i understand the 0.5. You can keep the 0.5 as long as you fixed the newton tests.

There was a problem with export of this new function `HRectangle` and also seems like with recent update statespace_sampler fails some other tests.

Datseris:
> and also seems like with recent update statespace_sampler fails some other tests.

please be specific. which tests? and why can you not fix them? are the tests here? are they in StateSpaceSets.jl? In any case you can do a PR to fix them wherever they are if you spotted the error.

StefanVaylBX2023:
There is a problem with new `statespace_sampler`, it is hard to understand your approach with threads but for some reason if you create a sampler function and will just do independent calls everything will be fine, but if you do smth like `[sampler() for _ in 1:1000]` it will overwrite the whole array to the same value, some of this things might be solved by doing `[copy(sampler()) for _ in 1:1000]` how i did in histogram_grouping test but ig thats not the way it is supposed to work (and it also passes this test 50/50 while the correct array of ics is generated, maybe something wrong with the seeding too).

Datseris:
no this is correct. `sampler()` over-writtes in place a memory array. If you need to _store_ the initial condition generated by the sampler, then do `u = copy(sampler())`. The functionality is like this because it is typically _not_ required to store the initial conditions generated by the sampler.

PR #86, opened 2023-07-21
Improve explanation of matching

Datseris:


Datseris:
I merge imeediatelly because this didn't havbe any code change

PR #87, opened 2023-07-28
use DocumenterCitations.jl / cite published article

Datseris:
This PR updates our citations to use our just-publushed chaos article.

It also switches the docbuild to DocumenterCitations.jl which seems like a great package. Let's hope it works!


Datseris:
docs looking very sexy!!! Thanks @goerz !

PR #88, opened 2023-08-03
Irregular grid

StefanVaylBX2023:
Initial implementation, but very strange behavior of mapper for default test of new implementation of regular grid, need to be discussed


Datseris:
i don't know what "very strange" means,  but I hope you will explain on our meeting today

Datseris:
? @StefanVaylBX2023 what happened here?

PR #90, opened 2023-08-04
Rematch, v2: a new dawn

Datseris:
This re-works the way matching works in `RAFM` to account for a _crucial mistake_. Previously, if an attractor dissapeared, and a new one appeared later, then they would both be assigned the same ID even though they would have never been compared with each other. 

This new continuation matching allows for two options:

1. Never compare vanished attractors, but still ensure that they get different ids
2. Compare vanished attractors by keeping track of their latest ghosts, and do the formal matching with the ghosts and new attractors

@kalelR this should be highly relevant for you if you use `RAFM` in scientific work.

@kalelR or @awage please review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v1.9
+- Matching attractors during the continuation with `RAFM` has been improved and is done by the function `match_continuation!` which has two options regarding how to handle attractors of previous parameters that have vanished.
+
 # v1.8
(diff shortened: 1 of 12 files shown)


KalelR:
Hey, excellent! Yup, I encountered this some time ago, and I fixed it with a similar idea to this `next_id` way, but with much hackier way, and didn't have the time to improve the code to submit a PR. 

This is very important, as you said. Glad it'll be fixed now! Thanks for the work :)

Issue #35, opened 2023-01-24
Basins of Attraction of irregularly spaced grid 


mdepitta:
How can I do a search/characterization of the basins of attraction on an irregular grid? It is, in fact, not an uncommon situation to have basins of attraction that are robust against some state variable perturbations rather than others. So it is convenient to have a search on, say, N points on variables that are known to have little impact on trajectories' ending points within a reasonable interval, but on 2xN or more points on those that are critical, and show high variability.       

Ideally say, we would like to have:

```using DynamicalSystems, OrdinaryDiffEq
  
    ds = ContinuousDynamicalSystem(my_model, u0, p)
    
    ## Say "my_model" has a state space of three variables 
    
    x1 = range(0,10,length=N_1)
    x2 = range(0,70,length=N_2)
    x3 = range(0.01,5,length=N_3)
    grid = (x1,x2,x3)     


    # # Calculate basins of attraction (old)
    reltol = 1e-6
    abstol = 1e-6
    diffeq = (alg = Vern9(), reltol = reltol, abstol = abstol)
    attractors = ... # Some assignment from known attractors of my_model 
    mapper = AttractorsViaProximity(mevmo, attractors; diffeq=diffeq)
    basins, = basins_of_attraction(mapper,grid)
```

When I tried to run something similar to this I get in `IJulia`:

MethodError: no method matching generate_ic_on_grid(::Tuple{Matrix{Float64}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, ::CartesianIndex{4})

Closest candidates are:
  generate_ic_on_grid(::Tuple{Vararg{T, B}}, ::Any) where {B, T} at ~/.julia/packages/ChaosTools/PHPDF/src/basins/attractor_mapping.jl:145

Datseris:
Notice that Attractors.jl is a separate module now that has several updates and modifications from the version of basins/tipping/etc. currently present in the feature frozen DynamicalSystems.jl documentation. I've transferred the issue to the appropriate repo. I'd recommend using Attractors.jl directly for now.

Datseris:
[...]

PR #91, opened 2023-08-09
Irregular grid support, issue #35

StefanVaylBX2023:


Datseris (commented):
Seems good. Perhaps you can use the example of Issue #35 as an example in the documentation?

StefanVaylBX2023:
> Seems good. Perhaps you can use the example of Issue #35 as an example in the documentation?

You mean docstring above the function or example for compiled documentation?

Datseris:
An example in the documentation. The the `docs` folder in the examples markdown file. Just like you did for the minimal fatal shock.

Datseris:
@StefanVaylBX2023 it is actually more useful to use this system here: https://github.com/JuliaDynamics/ChaosTools.jl/issues/258

Make the grid so that it is finer for x close to 0 and for y close to 0. It makes for the perfect example!

cc @KalelR  in case you still use fast slow systems.

StefanVaylBX2023:
> @StefanVaylBX2023 it is actually more useful to use this system here: [JuliaDynamics/ChaosTools.jl#258](https://github.com/JuliaDynamics/ChaosTools.jl/issues/258) as an example
> 
> Make the grid so that it is finer for x close to 0 and for y close to 0. It makes for the perfect example!
> 
> cc @KalelR in case you still use fast slow systems.

I've been running this example for different distributions of grid (you can control it with `pow` , the higher `pow`, the more finer grid you will get closer to 0), different `Dt`, different `u0`  and different total lengths of the grid. But I cannot understand results some of the results, can you help me figure out the behavior, if it is correct or not, here is some code I've been using:


```
using Attractors
using DynamicalSystems

@inbounds @inline function predator_prey(u, p, t)
	α, γ, ϵ, ν, h, K, m = p
	N, P = u
    du1 = α*N*(1 - N/K) - γ*N*P / (N+h)
    du2 = ϵ * (  ν*γ*N*P/(N+h) - m*P    )
	return SVector{2}(du1, du2)
end

γ = 2.5
h = 1 
ν = 0.5 
m = 0.4
ϵ = 1.0
α = 0.8 
K = 15 

u0 = rand(2)
ds = ContinuousDynamicalSystem(predator_prey, u0, [α, γ, ϵ, ν, h, K, m])
pow = 2
xg = yg = collect(range(0, 15^(1/pow), length=1000)).^ pow
mapper = AttractorsViaRecurrences(ds, (xg, yg);  Δt=0.1, sparse = false)
basins, attractors = basins_of_attraction(mapper; show_progress = true)

attractors[1]


using CairoMakie

heatmap_basins_attractors((xg,yg),basins, attractors)
```
and the results I obtained
plot_34

Datseris:
what do you not understand?

StefanVaylBX2023:
> what do you not understand?

I don't understand how this particular configuration is valid to be added as a final example.

Datseris:
wait , i am confused. when you run this code you do not find attractors...?

StefanVaylBX2023:
> wait , i am confused. when you run this code you do not find a limit cycle as attractors?

```
u0 = rand(2)
ds = ContinuousDynamicalSystem(predator_prey, u0, [α, γ, ϵ, ν, h, K, m])
pow = 4
xg = yg = collect(range(0, 15^(1/pow), length=10000)).^ pow
mapper = AttractorsViaRecurrences(ds, (xg, yg);  Δt=0.1, sparse = false)
mapper(u0)
u = mapper.bsn_nfo.attractors[1]
plot(u[:,1], u[:,2], seriestype = :scatter)
``` 
with this code i obtained a limit cycle, but it is not much different from results obtained in discussion

src/mapping/attractor_mapping_recurrences.jl
```diff
@@ -501,22 +534,41 @@ function relabel_visited_cell!(bsn_nfo::BasinsInfo, old_label, new_label)
 end
 
 function basin_cell_index(y_grid_state, bsn_nfo::BasinsInfo{B}) where {B}
```
Datseris, line 536: Please split this function into dwo functions that dispatch on the `grid_nfo` field.

Write `basin_cell_index(y, bsn_nfo) = basin_cell_index(y, bsn_nfo.grid_nfo)` and then use multiple dispatch to create two functions that dispatch on the grid type.

src/mapping/attractor_mapping_recurrences.jl
```diff
@@ -501,22 +534,41 @@ function relabel_visited_cell!(bsn_nfo::BasinsInfo, old_label, new_label)
...
+            return CartesianIndex{B}(-1)
+        end
+    else 
+        for (axis, coord) in zip(bsn_nfo.grid_nfo.grid, y_grid_state)
+            if coord < minimum(axis) || coord > maximum(axis)
+                return CartesianIndex{B}(-1)
+            end
+        end
+
+        cell_indices = map(x -> searchsortedlast(x[1], x[2]), zip(bsn_nfo.grid_nfo.grid, y_grid_state))
```
Datseris, line 563: don't write `x` and `x[1], x[2]`. Write `x, y` and use `x, y` in `searchsortedlast`.

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# v1.9.1
```
Datseris, line 1: As we are adding a new feature here, we need to increment the version to 1.10!

StefanVaylBX2023:
working on the comments rn, just one more question regarding example in the docs, what exactly do you expect to see there like plot of `mapper.bsn_info.attractors` or the heatmap.

Datseris:
I was thinking that a good example would be using this fast slow system, and solving it once with a _regular_ grid, and showing that it fails. Then solving it with hte _irregular_ grid and showing that it works (i.e., finds the limit cycle).

StefanVaylBX2023:
I've addressed all of the issues you mentioned but still problem with example. With `basins_of_attraction` for both cases all I can get is two saddle points as attractors (I've tried tons of initial conditions and grid setups). While running `mapper(u0)` with `u0 = rand(2)` writes down cyclical attractor into `bsn_nfo` again in both cases, with only difference that with the same setup irregular grid gives 20-30% more points

Datseris:
okay thanks, I'll have a look then!

PR #93, opened 2023-08-29
Irregular grids with discretization lvl specification 

StefanVaylBX2023:
2D works well 
3D has to be discussed, there is easier way to do that than via matrix, as construction of matrix by user in 3d is quite ambitious, one can just specify range of values which he desires to be more dense like some small cube (xg,yg,zg) and current algo will consume it with very few adjustments (also will work in higher dims if one would ever need this)
I will explain tomorrow the work of the algorithm, also I need some nice example for docs or should I try to reuse prays and predators?


Datseris:
yes, use predator pray as an example! and update the PR to current `main` branch so that thre are no conflucts!

Issue #84, opened 2023-07-04
Simpler and clearer grouping configuration for AttractorsViaFeaturizing


KalelR:
The standard way to group features in the `AttractorsViaFeaturizing` method is to clusterize them using DBSCAN. As we've known for some time, the algorithm is more or less a black box, and so it can be quite hard to debug it. Non-optimal parameters can lead to (i) finding the same attractor more than once (it might separate two trajectories on the same attractor that have slightly different feature values), or (ii) not finding one or more attractors that are there (so grouping 2+ attractors together). Sometimes both can occur for the same parametrization. In particular for the projects I've been working on, it has been *very* frustrating to figure out an optimal parametrization, as you really need to look deep into all details of the computations. 

As I understand, DBSCAN is great if you have noisy data, then it works great for identifying the clouds (clusters) of higher density. But in a wide range of applications I have in mind, this doesn't seem necessary, since 
1. The systems are deterministic
2. I know good features to separate the attractors 
3. I can run the trajectories for sufficiently long so that features on the same attractor differ very little.

As a consequence, feature space is composed of very small and dense clouds with a wide separation between themselves. DBSCAN is thus not really needed, we can use something much simpler / more intuitive. I've been having success with a simple comparison: just comparing the distances between each feature, and separating them if that distance is higher than a threshold. 

```
function _cluster_features_into_labels_comparing(features, config, distance_threshold; par_weight::Real = 0, plength::Int = 1, spp::Int = 1)
    labels_features = Vector{Int64}(undef, length(features)) #labels of all features
    labels_features[1] = 1
    clusters_idxs = [1] #idxs of the features that define each cluster
    cluster_labels = [1] #labels for the clusters, going from 1 : num_clusters
    next_cluster_label = 2
    metric = config.clust_distance_metric
    for idx_feature = 2:length(features)
        feature = features[idx_feature]
        dist_to_clusters = Dict(cluster_labels[idx] => evaluate(metric, feature, features[clusters_idxs[idx]]) for idx in eachindex(cluster_labels))
        dist_vals = collect(values(dist_to_clusters))
        min_dist, idx_min_dist = findmin(dist_vals)
        
        if min_dist > distance_threshold #bigger than threshold => new attractor
            push!(clusters_idxs, idx_feature)
            push!(cluster_labels, next_cluster_label)
            feature_label = next_cluster_label
            next_cluster_label += 1
        else #smaller => assign to closest cluster 
[...]

PR #97, opened 2023-09-30
Add new grouping method: AttractorsViaPairwiseComparison

KalelR:
As we had discussed in Issue #84, this PR adds a simpler grouping method to "clusterize" features, particularly useful of course when using `AttractorsViaFeaturizing`.

The idea is that clustering algorithms aren't really necessary for grouping features. The systems are deterministic, and we can usually integrate systems long enough to have the features converge. If the user knows good features, and for how to integrate them, then the features in feature space will be very well behaved, with attractors corresponding to small clouds well separated from each other. Therefore, there is no need to use clustering algorithms. Attractors can be found directly by simply grouping together features close to each other within some epsilon, and separating those farther away.

The algorithm is thus very simple. Basically
1.    Assign feature 1 as a new attractor -> clusters_idxs = [1]
2. For each subsequent feature,
       1. compute distance from the future to the previously found attractors -> dist(feature, clusters_idxs)
        2. find minimum distance between future and cluster idxs (ie find the the cluster that is closest to the current feature)
        if min dist > threshold, consider the future a new attractor; if min dist < threshold, assign it to the closest cluster

This is nicer than simple histograms mainly because it allows for more general `features`, beyond just vectors of numbers. They can be, for instance, matrices, which allows for features to be the entire attractors. The distance metric just needs to take in `feature_A` and `feature_B` and return a number. Plus, given that features are well behaved, the `distance_threshold` doesn't really need to be fined-tuned much. This is wayy more reliable, understandable, and easier to use than DBSCAN. Since I've implemented this I never went back. And to be honest I don't see any future scenario when I would.

One part left to do, if needed, is to implement support for computing distances using these parameters:
```
par_weight::Real = 0, plength::Int = 1, spp::Int = 1

### DIFF
--- a/src/mapping/grouping/pairwise_comparison.jl
+++ b/src/mapping/grouping/pairwise_comparison.jl
@@ -22,11 +22,16 @@ features.
   separately into the range `[0,1]`. This typically leads to more accurate clustering.
 
 ## Description
-
-This algorithm considers that features belong to the same attractor when their pairwise
-distance, computed using `distance_metric` is smaller or equal than `distance_threshold`,
-and belong to different attractors when the distance is bigger than the threshold.
-Attractors correspond to each grouping of similar features.  
+This algorithm assumes that the the features well-separated into distinct clouds, with the
+maximum radius of the cloud controlled by `distance_threshold`. Since the systems are
+deterministic, this is achievable with a good-enough `featurizer` function, by removing
+transients, and running the trajectories for sufficiently long. It then considers that
+features belong to the same attractor when their pairwise distance, computed using
+`distance_metric`, is smaller or equal than `distance_threshold`, and belong to different
+attractors when the distance is bigger than the threshold. Attractors correspond to each
+grouping of similar features. In this way, the key parameter `distance_threshold` is
+simply the amount of variations permissible in the features. If they are well-chosen, the
+value can be relatively small and does not need to be fine tuned.
 """
 struct GroupViaPairwiseComparison{R<:Real, M} <: GroupingConfig
     optimal_radius_method::R
@@ -36,11 +41,11 @@ end
 
 function GroupViaPairwiseComparison(;
         optimal_radius_method, #impossible to set a good default value, depends on the features
-        clust_distance_metric=Euclidean(), rescale_features=false, 
+        distance_metric=Euclidean(), rescale_features=false, 
     )
     return GroupViaPairwiseComparison(
         optimal_radius_method,
-        clust_distance_metric, rescale_features,
+        distance_metric, rescale_features,
     )
 end
 
@@ -71,16 +76,18 @@ function _cluster_features_into_labels(features, config::GroupViaPairwiseCompari
     cluster_labels = [1] # labels for the clusters, going from 1 : num_clusters
     next_cluster_label = 2
     
-    for (idx_feature, feature) in enumerate(features[2:end])
-        dist_to_clusters = Dict(cluster_label => evaluate(metric, feature, features[cluster_idxs[idx_cluster]]) for (idx_cluster, cluster_label) in enumerate(cluster_labels))
+    for idx_feature = 2:length(features)
+        feature = features[idx_feature]
+        dist_to_clusters = _distance_dict(feature, features, cluster_idxs, cluster_labels, metric)
         min_dist, closest_cluster_label = findmin(dist_to_clusters)
         
         if min_dist > distance_threshold #bigger than threshold => new attractor
+        @show feature, features[cluster_idxs]
             feature_label = next_cluster_label
             push!(cluster_idxs, idx_feature)
             push!(cluster_labels, next_cluster_label)
+            # @info "New attractor $next_cluster_label, min dist was $min_dist > $distance_threshold" #TODO: allow this when debugging verbose mode on!
             next_cluster_label += 1
-            @info "New attractor, min dist was $min_dist > $distance_threshold"
         else #smaller than threshold => assign to closest cluster 
             feature_label = closest_cluster_label
         end
@@ -88,4 +95,14 @@ function _cluster_features_into_labels(features, config::GroupViaPairwiseCompari
         labels_features[idx_feature] = feature_label 
     end 
     return labels_features
+end
+
+
+function _distance_dict(feature, features, cluster_idxs, cluster_labels, metric)
+    if metric isa Metric
+        dist_to_clusters = Dict(cluster_label => evaluate(metric, feature, features[cluster_idxs[idx_cluster]]) for (idx_cluster, cluster_label) in enumerate(cluster_labels))
+    else
+        dist_to_clusters = Dict(cluster_label => metric(feature, features[cluster_idxs[idx_cluster]]) for (idx_cluster, cluster_label) in enumerate(cluster_labels))
+    end
+    return dist_to_clusters 
 end
--- a/src/mapping/grouping/pairwise_comparison.jl
+++ b/src/mapping/grouping/pairwise_comparison.jl
@@ -1,0 +1,91 @@
+export GroupViaPairwiseComparison
+
+
+"""
+    GroupViaPairwiseComparison(; optimal_radius_method::Real, kwargs...)
+
+Initialize a struct that contains instructions on how to group features in
+[`AttractorsViaFeaturizing`](@ref). `GroupViaPairwiseComparison` groups features and
+identifies clusters by considering the pairwise distance (using `distance_metric`) between
+features. 
+
+## Keyword arguments
+    
+* `distance_threshold`: a real number defining the maximum distance two features can be to
+  be considered in the same cluster - above the threshold, features are different. This
+  value simply needs to be large enough to differentiate clusters.
+* `distance_metric = Euclidean()`: A metric to be used in the clustering. It can be any
+  function `f(a, b)` that returns the distance between any type of data structure (usually
+  vectors or matrices of reals). Needs to be consistent with the `featurizer` function. All
+  metrics from Distances.jl can be used here.
+* `rescale_features = true`: if true, rescale each dimension of the extracted features
+  separately into the range `[0,1]`. This typically leads to more accurate clustering.
+
+## Description
+
+This algorithm considers that features belong to the same attractor when their pairwise
+distance, computed using `distance_metric` is smaller or equal than `distance_threshold`,
+and belong to different attractors when the distance is bigger than the threshold.
+Attractors correspond to each grouping of similar features.  
+"""
+struct GroupViaPairwiseComparison{R<:Real, M} <: GroupingConfig
+    optimal_radius_method::R
+    distance_metric::M
+    rescale_features::Bool 
+end
+
+function GroupViaPairwiseComparison(;
+        optimal_radius_method, #impossible to set a good default value, depends on the features
+        clust_distance_metric=Euclidean(), rescale_features=false, 
+    )
+    return GroupViaPairwiseComparison(
+        optimal_radius_method,
+        clust_distance_metric, rescale_features,
+    )
+end
+
+function group_features(
+    features, config::GroupViaPairwiseComparison;
+    kwargs...
+)
+    if config.rescale_features
+        features = _rescale_to_01(features)
+    end
+    ϵ_optimal = _extract_ϵ_optimal(features, config)
+    labels = _cluster_features_into_labels(features, config, ϵ_optimal; kwargs...)
+    return labels
+end
+
+function _extract_ϵ_optimal(features, config::GroupViaPairwiseComparison)
+    return config.optimal_radius_method
+end
+
+# TODO: add support for par_weight,plength and spp?
+function _cluster_features_into_labels(features, config::GroupViaPairwiseComparison, distance_threshold::Real; par_weight::Real = 0, plength::Int = 1, spp::Int = 1)
+    labels_features = Vector{Int64}(undef, length(features)) #labels of all features
+    metric = config.distance_metric
+    
+    # Assign feature 1 as a new attractor
+    labels_features[1] = 1
+    cluster_idxs = [1] # idxs of the features that define each cluster
+    cluster_labels = [1] # labels for the clusters, going from 1 : num_clusters
+    next_cluster_label = 2
+    
+    for (idx_feature, feature) in enumerate(features[2:end])
+        dist_to_clusters = Dict(cluster_label => evaluate(metric, feature, features[cluster_idxs[idx_cluster]]) for (idx_cluster, cluster_label) in enumerate(cluster_labels))
+        min_dist, closest_cluster_label = findmin(dist_to_clusters)
+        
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/mapping/grouping/pairwise_comparison.jl [hunk 1]
```diff
@@ -0,0 +1,108 @@
...
+        features = _rescale_to_01(features)
+    end
+    ϵ_optimal = _extract_ϵ_optimal(features, config)
+    labels = _cluster_features_into_labels(features, config, ϵ_optimal; kwargs...)
+    return labels
+end
+
+function _extract_ϵ_optimal(features, config::GroupViaPairwiseComparison)
+    return config.optimal_radius_method
+end
```
Datseris, line 66: I am confused about the source code here... Why do you need this function? And why is something called `ε_optimal` at some places and `distance_threshold` at others? Are we doing any clustering here after all? No right? There isn't any `ε_optimal`? Or is there. :S

Datseris:
What do you mean that you need to add support for 
```
par_weight::Real = 0, plength::Int = 1, spp::Int = 1
```
?

Is this for the continuation? If so, `plength, spp` come from the continuation functions, in which sense do you need to add support for them?

`par_weight` is something we can, and should, safely ignore. I can't conceive a **scientific** argument why anyone should ever use `par_weight`. We had it in the first clustering-continuation because it is in the MCBB paper. Whether it makes sense... it doesn't to me.

Datseris:
it is not clear to me how one would choose the distance threshold. I guess with the scaling to 0-1 turned on, a good value could be 0.1 or something like that?

Datseris:
Is this method **significantly faster** than the clustering method? You should do formal benchmarks and post the results here. Compute a list of features, and then benchmark the `group_features` function with both algorithms.

test/mapping/attractor_mapping.jl
```diff
@@ -218,9 +245,9 @@ if DO_EXTENSIVE_TESTS
             g = exp(entropy(Renyi(; q = 0), probs))
             return SVector(g, minimum(A[:,1]))
         end
-
+        
         test_basins(ds, u0s, grid, expected_fs_raw, featurizer;
-        ε = 0.01, ferr=1e-2, Δt = 0.2, mx_chk_att = 20)
+        ε = 0.01, ferr=1e-2, Δt = 0.2, mx_chk_att = 20, distance_threshold_pairwise=100) #threshold is very high because features haven't really converged yet here
```
Datseris, line 250: How can the threshold be larger than 1? If we scale to 0-1 by default, the distance in feature space can never be more than the sqrt of the number of feature dimensions.

Datseris: Ah. we dont rescale be default. I think we should.

src/mapping/grouping/pairwise_comparison.jl
KalelR: That was stupid of me, I'll change everything to `distance_threshold` as it should be.

KalelR:
> What do you mean that you need to add support for
> 
> ```
> par_weight::Real = 0, plength::Int = 1, spp::Int = 1
> ```
> 
> ?
> 
> Is this for the continuation? If so, `plength, spp` come from the continuation functions, in which sense do you need to add support for them?
> 
> `par_weight` is something we can, and should, safely ignore. I can't conceive a **scientific** argument why anyone should ever use `par_weight`. We had it in the first clustering-continuation because it is in the MCBB paper. Whether it makes sense... it doesn't to me.



As I understand they included this parameter distance in the computation of distances for the clustering. The idea was to "ensure that similar asymptotic states with strongly different parameter values are distinguished from each other", which I never really understood to be honest haha - isn't the whole point of continuation tracking an attractor across different parameters?. But anyways, I thought that if people wanted this in the continuation method via clustering, they might also want it in the continuation method using this PairwiseComparison - the latter doesn't cluster per se, but we still computes distances. What do you think?

TLDR: This can be easily implemented in the future if anyone wishes. For now, I don't think it's needed, but wanted to check with you.

KalelR:
> it is not clear to me how one would choose the distance threshold. I guess with the scaling to 0-1 turned on, a good value could be 0.1 or something like that?

The choice is a balance between two points
1. you want it to be large enough to account for variations in the features from the same attractor - if it's too small, the algorithm will find duplicate attractors (which to be fair is easy to deal with, and all other algorithms also have this problem).
2. you want it to be small enough to not group together features from distinct attractors. This requires some knowledge of how spread your features are. If it's too big, the algorithm will miss some attractors, as it groups 2+ distinct attractors. This is harder to deal with.

So what I've been doing is to ensure the deviations in the features are super small, which can be achieved with a good choice of `featurizer`, `T` and `Ttr. Then I set `distance_threshold` to a small value, like 1e-2 or 1e-3 even for non-rescaled features. I reduce it until no more unique attractors are found, and then just ensure that there are no duplicate attractors.

test/mapping/attractor_mapping.jl
KalelR: I don't think we should, there are some serious problems with rescaling the features, and the user might get lost debugging the algorithm until they realize what the rescaling did with the features. We discussed this a bit in PR #59 but actually there's more to it I think. I will write my ideas about this down soon and then we can see.

For now I defend that rescaling should be done with a lot of care, so it should be disabled by default.

KalelR:
> Is this method **significantly faster** than the clustering method? You should do formal benchmarks and post the results here. Compute a list of features, and then benchmark the `group_features` function with both algorithms.

Not sure if it's significantly faster, but I don't think the speed really matters to be honest. It's so much simpler, more reliable and easy to use than the clustering methods that I wouldn't mind even if it wasn't slower. But now, I think at worse it's a comparable speed. DBSCAN at least requires the computation of the distance matrix, which is order `~N^2`. This method, the way I implemented at least, does `~N*number_attractors`, which is generally smaller. Memory-wise it's much better also, as it only stores `~number_attractors` distances for each iteration.

Datseris:
> It's so much simpler, more reliable and easy to use than the clustering methods that I wouldn't mind even if it wasn't slower.

I don't agree with this statement. I agree that it is simpler, but I can't see a way that it is more reliable. You need to pick this distance number. In the clustering you don't pick anything, it is done automatically.

Not only that, but the quality of your clustering depends on the _order_ of the initial conditions. Hence, each time you call `basins_fractions` you will get different results unless you start the simulation with the same seed. It's because the "center" of a cluster in this new algorithm is the very first point, and the very first point is random.

DBSCAN, as well as all other grouping algorithms we have so far, are invariant to the sequence of the initial conditions.

Note that this doesn't stop us from merging this in, but I am just pointing out that I personally wouldn't use this one instead of the Clustering version. If I knew the system so well that I knew what the features would do, this means that I know the attractors. Then, I would just use `AttractorsViaProximity` which is the least ambiguous out of all 4 methods.

Datseris:
> The idea was to "ensure that similar asymptotic states with strongly different parameter values are distinguished from each other", which I never really understood to be honest haha - isn't the whole point of continuation tracking an attractor across different parameters?.

Exactly. Hence, I would argue, this whole `par_weight` business is a bad choice. My experience working on "Framework for global stability" leads me to this conclusion. I say we should ignore this `par_weight` completely. Perhaps we should even hide it from the Clustering continuation in the future documentation as well. This wouldn't be a breaking change, but we simply would no longer support it.

Datseris:
> Not sure if it's significantly faster, but I don't think the speed really matters to be honest

Sure but please do a formal benchmark and report the results. It shouldn't take too much effort and performance matters in this context. A lot. For me, it is the prime reason to change algorithms. Otherwise, I would never use anything else beyond the Recurrences method.

Datseris:
You should document tips for choosing the distance threshold by summarizing what you wrote her.

Datseris:
> that the attractors themselves could be used as features. Then one would separate attractors by some distance metric in state space directly

You would need to find the attractors first. The only method we have that finds attractors is the recurrences one. And if you use this method, you need to other method to group attractors, as the recurrences anyways maps individual i.c. to attractors correctly.

---

> DBSCAN also requires a distance parameter, which is much less intuitive (leads to the black box behavior we discussed so many times). The only reason the clustering method as a whole doesn't require this parameter is that we brute-force search for an "optimal value". This is very debatable. In the use cases I've tried (on a system of 2 coupled 2d oscillators and on 100d networks) this lead to problems so often. Sometimes the "optimal" parameter was way too small, and the algorithm found lots of duplicates. Sometimes it was way too big, and missed some attractors. Sometimes it worked great for one system parameter and then terrible for another with similar dynamics.
> 
> This simple comparison came as a way to simplify the procedure and enable me some clarity to know when the algorithm really is working. This is cases when I don't know the attractors of the system, but I can probe around, change parameters and `featurizer` until I'm confident the correct attractors were found (no duplicates, no over-grouping).


That's fine, but then you should somehow summarize this discussion in the docstring of the new method for when it should be preferred over the Clustering method.

---

> If the worry is having an automatic procedure, we could think of a search algorithm here as well. Something like reducing `distance_threshold` until no more unique attractors are found and no duplicates appear.

Given the amount of papers on optimizng the clustering of DBSCAN, and the fact that to this day we even presented an even more optimal way, this could take a lot of time. You should open an Issue Request for this after we merge this PR in.

KalelR:
I benchmarked the two methods using two sets of features: one that is just 5 randomly generated features repeated N times, another extracted from Henon (like in the tests). The distance threshold was the same for both, and they find the correct clusters. As expected from the difference in scaling, this pairwise comparison is faster and uses less memory than DBSCAN. For some reason in the Henon case the number of allocations in DBSCAN is smaller, but the memory usage is bigger. I don't know why.

It is somewhat unfair to compare them though, as the pairwise comparison is so much simpler than DBSCAN.

display

Code is here https://gist.github.com/KalelR/1c3a66a157a8e347dcd6d6ad28f633cc, also in our FrameworkGlobalStability repo.

Datseris:
Great, thanks, you should state the difference in performance scaling in the docstring. 

Please also address the following comment in the docstring:


> DBSCAN also requires a distance parameter, which is much less intuitive (leads to the black box behavior we discussed so many times). The only reason the clustering method as a whole doesn't require this parameter is that we brute-force search for an "optimal value". This is very debatable. In the use cases I've tried (on a system of 2 coupled 2d oscillators and on 100d networks) this lead to problems so often. Sometimes the "optimal" parameter was way too small, and the algorithm found lots of duplicates. Sometimes it was way too big, and missed some attractors. Sometimes it worked great for one system parameter and then terrible for another with similar dynamics.
> 
> This simple comparison came as a way to simplify the procedure and enable me some clarity to know when the algorithm really is working. This is cases when I don't know the attractors of the system, but I can probe around, change parameters and `featurizer` until I'm confident the correct attractors were found (no duplicates, no over-grouping).


(That's fine, but then you should somehow summarize this discussion in the docstring of the new method for when it should be preferred over the Clustering method.)

---

Lastly, please add an example in the examples list that utilizes this method. Would be fantastic if you used a real world example, i.e., something from your own work, as you said that this method helped you there a lot.

---

Then, I can merge!

(and don't forget to increment minor version and add changelog entry)

KalelR:
> > that the attractors themselves could be used as features. Then one would separate attractors by some distance metric in state space directly
> 
> You would need to find the attractors first. The only method we have that finds attractors is the recurrences one. And if you use this method, you need no other method to group attractors, as the recurrences anyways maps individual i.c. to attractors correctly.
> 
I misspoke, I meant that the _trajectories_ can be used as features themselves. Meaning the featurizer function `f(A, t)` can simply return `A` and the algorithm will be fine with this as long as the distance metric takes in `A` and `B`, two trajectories, and returns some value.

KalelR:
> Lastly, please add an example in the examples list that utilizes this method. Would be fantastic if you used a real world example, i.e., something from your own work, as you said that this method helped you there a lot.

This would be quite nice, I'll think of a nice example where it helped. But since the work is still unpublished, I'd rather wait a month or so before putting this online with all the methodology and parameters correctly chosen, just to be sure. Is that alright?

src/mapping/grouping/pairwise_comparison.jl
```diff
@@ -0,0 +1,116 @@
...
+
+## Keyword arguments
+    
+* `distance_threshold`: a real number defining the maximum distance two features can be to
+  be considered in the same cluster - above the threshold, features are different. This
+  value simply needs to be large enough to differentiate clusters.
+* `distance_metric = Euclidean()`: A metric to be used in the clustering. It can be any
+  function `f(a, b)` that returns the distance between any type of data structure (usually
+  vectors or matrices of reals). Needs to be consistent with the `featurizer` function. All
+  metrics from Distances.jl can be used here.
```
Datseris, line 20: We can simplify here. First, let's do a mass-rename of `distance_metric` to simply `metric`. This is the name we use throughout the ecosystem. Then, let's do a mass rename of `distance_threshold` to just `threshold`, as similarly, that's the name we use. There is no reason to ambiguate with `distance` in front, because there is no other threshold in this functionality. Then, let's simply state as a description:

"A _function_ that given two features `metric(a, b)` it returns the distance between them. `a, b` are outputs of the `featurizer` function. Any `Metric` from Distances.jl may be used as well."

src/mapping/grouping/pairwise_comparison.jl
```diff
@@ -0,0 +1,116 @@
...
+* `rescale_features = true`: if true, rescale each dimension of the extracted features
```
Datseris, line 21: You argued that this should be false but here it is true (and you also say it leads to more accurate clustering)?

src/mapping/grouping/pairwise_comparison.jl
```diff
@@ -0,0 +1,116 @@
...
+[`AttractorsViaFeaturizing`](@ref). `GroupViaPairwiseComparison` groups features and
+identifies clusters by considering the pairwise distance between features. It can be used
+as an alternative to the clustering method in `GroupViaClustering`, having the
+advantage that it is simpler, typically faster and uses less memory. 
+
+## Keyword arguments
+    
+* `distance_threshold`: a real number defining the maximum distance two features can be to
+  be considered in the same cluster - above the threshold, features are different. This
+  value simply needs to be large enough to differentiate clusters.
```
Datseris, line 16: before the description say something like (mandatory keyword).

Datseris (commented):
Sorry only minor comments and then I'll merge!

src/mapping/grouping/pairwise_comparison.jl
KalelR: I kept the same text as in the the clustering method haha. We can keep it like this, so the texts are consistent, and then discuss if the rescaling is appropriate or not separately. I also realized I exaggerated before, rescaling is actually fine. The biggest problem I had with rescaling doesn't happen now, because we are rescaling differently than before, when I had problems with it. There's only a minor case we need to deal with, I'll open a PR about this.

Datseris (approved):

Datseris:
We don't have to rescale to [0, 1]. We can use `StateSpaceSets.standardize` instead, which rescales to 0 mean and 1 standard deviation.

Issue #76, opened 2023-06-06
Creating ics on grid for basins of attraction is super slow for higher dimensions


KalelR:
Hey

I need to calculate basins of attraction for a high-dimensional system (100+), but I noticed that `basins_of_attraction` is super slow for these cases because of the creation of the initial conditions on a grid. Especifically
`A = StateSpaceSet([generate_ic_on_grid(grid, i) for i in vec(I)])`

is very slow for higher dimensions (was taking more than an hour to run...). 

I think there are two reasons: evaluating the `CartesianIndices` becomes slow, and also maybe because `generate_ic_on_grid` allocates StaticVectors.   

What would be a good solution for this?

Thanks a lot!

awage:
I wouldn't use `basins_of_attraction` for this purpose. Use directly the `mapper`: 

```julia
lab = zeros(Int32, N)
for i in 1:N
      lab[i] = mapper( ic_generator())
end
```
The `ic_generator` can be a function or an array on the region you want to explore. The memory consumption would be restricted to the sparse matrix. For Kuramotos it works well. 

A more condensend version: `lab = [ mapper(ic_gen()) for i in 1:N]`

KalelR:
Ah, nice idea! But in this case I need to use `AttractorsViaFeaturizing` since the dimension is so high.

awage:
So the problem is just initial condition caching? Because I assume you have a low dimensional feature space.

KalelR:
> So the problem is just initial condition caching?

Yup. This line
`A = StateSpaceSet([generate_ic_on_grid(grid, i) for i in vec(I)])`

takes way too long just to create the ics when the dimension of the grid is high (100+). 
I could of course just create `A` how I want for my specific problem, but it would be great if there was a simple and general way to create the ics that also works well in higher dims.  


 > Because I assume you have a low dimensional feature space.

Yup, `basins_fractions` works great once we have the ics.

awage:
What about a special sampler with an internal queue: 

`sampler() =  generate_ic_on_grid(grid, pop!(I))`

I am not sure it is possible, I am just throwing an idea

Datseris:
It is because of static arrays.

We need to allow `StateSpaceSet` to work with normal vectors as well. Someone needs to do a PR in StateSpaceSets.jl.

Datseris:
[...]

PR #99, opened 2023-10-01
Creating ics on grid for higher dimensions

KalelR:
This is the code I used to deal with the problem we discussed in Issue #76. There, using `basins_of_attraction` in high (maybe >5?) dimensional-systems wouldn't run, because the algorithm would get stuck generating initial conditions. I explain the idea for my solution in Issue #76. It's probably not the most optimal way of solving the problem, but it works and is fast enough for my use-cases. Let me know if you think it's worth keeping it like this!

### DIFF
--- a/src/mapping/attractor_mapping.jl
+++ b/src/mapping/attractor_mapping.jl
@@ -1,6 +1,8 @@
 # Definition of the attracting mapping API and exporting
 # At the end it also includes all files related to mapping
 
+const mx_dimension_sparse = 30
+
 export AttractorMapper,
     AttractorsViaRecurrences,
     AttractorsViaProximity,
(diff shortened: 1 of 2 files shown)


Datseris:
I don't understand this PR. Can you please explain it? What are dimensions that are not "varying" ? What does this mean?

If someone calls `basins_of_attraction(mapper, grid)` then naturally all dimensions of the grid are varying, right? Or? This PR adds a function `generate_ics_on_grid` but doesn't reference this function anywhere, so I am not sure where or how this function would be used. Can you give a concrete example application?

Datseris:
bump @KalelR

Datseris:
bump @KalelR

Datseris:
@KalelR I'm closing this, as it hasn't received the necessary attention to progress. If you want to work on this you can open it again. Stale PRs make a repo look "abandoned" to some people.

PR #101, opened 2023-10-01
Cleanup and clarify recurrences src / Fix #100

### LINKED ISSUE #100: Why do we store the visited cells...?
cc @awage 

Why does `BasinInfo` contain a field `visited_list::Vector{CartesianIndex{D}}`? Is this used anywhere? I am reading the source code but I don't understand if it is used anywhere.

THe only point in the source code that `.visited_list` is populated with cells is here:

```julia
    if bsn_nfo.state == :att_search
        if ic_label == 0
            # unlabeled box, label it with current odd label and reset counter
            bsn_nfo.basins[n] = bsn_nfo.visited_cell_label
            push!(bsn_nfo.visited_list, n) # keep track of visited cells
            bsn_nfo.consecutive_match = 1
        elseif ic_label == bsn_nfo.visited_cell_label
            # hit a previously visited box with the current label, possible attractor?
            bsn_nfo.consecutive_match += 1
        end
```

There are only 2 sections in the source code where `.visited_list` is called, and they are both the same function call:

```julia
relabel_visited_cell!(bsn_nfo, bsn_nfo.visited_cell_label, 0)
```

which calls the function:

```julia
function relabel_visited_cell!(bsn_nfo::BasinsInfo, old_label, new_label)
    while !isempty(bsn_nfo.visited_list)
        ind = pop!(bsn_nfo.visited_list)
        if bsn_nfo.basins[ind] == old_label
            bsn_nfo.basins[ind] = new_label
        end
    end
end
```

There are two things I notice:

1. `old_label` **can never be the same as** `new_label`, as the visited cells by definition have label at least 4, while `new_label` is always `0` given how we call the function.
2. We **do not use the list of visited cells**. Like, wow. The way we check if we have entered a visited cell is by checking the **cell label** that is stored in the basins array.

So, @awage, do we actually need the `visited_list` ? I don't see how. Seems to me like we are wasting computations by storing, and then removing the visited cells, without actually using them anywhere!

Datseris:
I am working on functionality that given a dynamical system it animates the action of `AttractorsViaRecurrences`. I will contribute this in a separate PR, but I am doing pretty good, here is where I am now: 

https://github.com/JuliaDynamics/Attractors.jl/assets/19669089/d4e856fe-85da-473d-b00b-10f8211c9740

During this journey, I had some problems with our source code. I admit, I find it surprising how difficult the source code is to read, given the success of the algorithm.

In this PR I do the following:

- Re-organize the source code more logically
- Re-named many variables and functions so that they better convey what they do. E.g., `visited_cell` was a poor name. It contained the (default) label that visited cells obtain. You'd think it would contain a cell (i.e., a cartesian index instead). A better name is `visited_cell_label`. Etc.
- Fix #100 
- Add many more comments to the source code to better explain things


Datseris:
oooof oh boy this is finally done. Took me quite a while to go through the source code and really clarify things as much as possible.

@awage or @kalel can you please review and leave an Approving review if you don't find anything wrong so that I can merge this in? This PR does not alter any functionality, it only re-wrote the source code to be more readable.

awage:
Hi! I will review the code today and leave some comments if any.

src/mapping/recurrences/attractor_mapping_recurrences.jl
```diff
@@ -0,0 +1,326 @@
...
+  which may be desirable for instance in chaotic attractors.
+* `mx_chk_lost = 20`: Maximum check of iterations outside the defined grid before we
+  declare the orbit lost outside and hence assign it label `-1`.
+* `horizon_limit = 1e6`: If the norm of the integrator state reaches this
+  limit we declare that the orbit diverged to infinity.
+* `mx_chk_safety = Int(1e6)`: A safety counter that is always increasing for
+  each initial condition. Once exceeded, the algorithm assigns `-1` and throws a warning.
+  This clause exists to stop the algorithm never haulting for innappropriate grids,
+  where a found attractor may intersect in the same cell with a new attractor the orbit
+  traces (which leads to infinite resetting of all counters).
```
awage, line 81:
```suggestion
This clause exists to stop the algorithm when it never halts due to inappropriate grids. For example it happens when the orbit of a newly found attractor intersect in the same cell of a previously found attractor  (which leads to infinite resetting of all counters) ```

File: src/mapping/recurrences/attractor_mapping_recurrences.jl [hunk 2]
```diff
@@ -0,0 +1,326 @@
...
+    Δt::Δ
+    state::Symbol
+    current_att_label::Int
+    visited_cell_label::Int
+    consecutive_match::Int
+    consecutive_lost::Int
+    prev_label::Int
+    safety_counter::Int
+    attractors::Dict{Int, StateSpaceSet{D, T}}
+    visited_list::Vector{CartesianIndex{D}}
```
awage, line 236: `visited_list` can be renamed as `visited_cells` or `visited_cells_list`. Technically it is not a list, it is a stack.

src/mapping/recurrences/attractor_mapping_recurrences.jl
awage, line 81:
```suggestion
  This clause exists to stop the algorithm never halting for inappropriate grids. It may happen 
  when a newly found attractor orbit intersects in the same cell of a previously found attractor (which leads to infinite resetting of all counters).
```

src/mapping/recurrences/finite_state_machine.jl
```diff
@@ -0,0 +1,275 @@
...
+
+
+function reset_basins_counters!(bsn_nfo::BasinsInfo)
+    bsn_nfo.consecutive_match = 0
+    bsn_nfo.consecutive_lost = 0
+    bsn_nfo.prev_label = 0
+    bsn_nfo.state = :att_search
+end
+
+function update_fsm_state!(bsn_nfo, ic_label)
```
awage, line 240: This function is very cryptic. I am sure there is a better way to express the transitions from state to state. I will try to improve this part.

src/mapping/recurrences/attractor_mapping_recurrences.jl
```diff
@@ -0,0 +1,326 @@
...
+    ind = iseven.(basins)
+    basins[ind] .+= 1
+    basins .= (basins .- 1) .÷ 2
+    return basins, mapper.bsn_nfo.attractors
+end
+
+#####################################################################################
+# Definition of `BasinInfo` and initialization
+#####################################################################################
+abstract type Grid end
```
awage, line 222: This type should be in grids.jl, there is a reason to declare this here?

src/mapping/recurrences/grids.jl
```diff
@@ -0,0 +1,169 @@
...
+    steps = map(r -> mean(diff(r)), g)
+    return norm(steps)
+end
+
+"""
+    SubdivisionBasedGrid(grid::NTuple{D, <:AbstractRange}, lvl_array::Array{Int, D})
+
+Given a coarse `grid` tesselating the state space, construct a `SubdivisionBasedGrid`
+based on the given level array `lvl_array`.
+The level array is has non-negative integer values.
```
awage, line 37:
```suggestion
The level array has non-negative integer values.
```

src/mapping/recurrences/grids.jl
```diff
@@ -0,0 +1,169 @@
...
+Value 0 means that the corresponding cell of the coarse `grid` is not subdivided
+any further. Value `n > 0` means that the corresponding cell will be subdivided
+in total `2^n` times (along each dimension), resulting in finer cells
+within the original coarse cell.
```
awage, line 41:
```suggestion
Given a coarse `grid` tesselating the state space, construct a `SubdivisionBasedGrid`
based on the given level array `lvl_array` that should have the same dimension as `grid`. 
The level array has non-negative integer values, with 0 meaning that the 
corresponding cell of the coarse `grid` should not be subdivided any further. 
Value `n > 0` means that the corresponding cell will be subdivided
in total `2^n` times (along each dimension), resulting in finer cells
within the original coarse cell.
```

src/mapping/recurrences/grids.jl
```diff
@@ -0,0 +1,169 @@
...
+        if (isequal(norm(velocity),NaN))
+            velocities[ind] = Inf
+        else
+            velocities[ind] = norm(velocity)
+        end
+    end
+
+    maxvel = maximum(filter(x -> x != Inf, velocities))
+    velratios = maxvel./velocities
+    result = [round(Int,log2(clamp(x, 1, 2^maxlevel))) for x in velratios]
```
awage, line 124: This procedure is documented somewhere? What is the criterion for the subdivision? Why the normalization of the velocities is upside down? I would have expected `velratios = velocities/mavel`

src/mapping/recurrences/grids.jl
awage: I am not sure about the dimension of the grid. You can correct this.  

Another point: this is the documentation of a type. It should be explicit in the docstring?

awage:
I am done with the review. There is the function `update_fsm_state!` that should be changed. I will think about a better way to write this function.

Datseris:
Thanks @awage. I have addressed all comments. Please leave an _approving_ review (by clicking review, and then selecting approve), if you think this should be merged.

awage (approved):

PR #103, opened 2023-10-08
Make `AttractorsViaRecurrences` more simple when finding and storing attractors

### LINKED ISSUE #102: AttractorsViaRecurrences `mx_chk_loc_att` and `:att_found` state
Alright, @awage I think we should discuss this. At the moment, the mode of operation `:att_found` of the Finite State Machine (FSM) is as follows:

```julia
    # This state can only be reached from `:att_search`. It means we have
    # enough recurrences to claim we have found an attractor.
    # We then locate the attractor by recording enough cells.
    if bsn_nfo.state == :att_found
        if ic_label == 0 || ic_label == bsn_nfo.visited_cell_label
            # label this cell as part of an attractor
            bsn_nfo.basins[n] = bsn_nfo.current_att_label
            bsn_nfo.consecutive_match = 1
            store_attractor!(bsn_nfo, u)
        elseif iseven(ic_label) && (bsn_nfo.consecutive_match <  mx_chk_loc_att)
            # We make sure we hit the attractor another `mx_chk_loc_att` consecutive times
            # just to be sure that we have the complete attractor
            bsn_nfo.consecutive_match += 1
            store_once_per_cell || store_attractor!(bsn_nfo, u)
        elseif iseven(ic_label) && bsn_nfo.consecutive_match ≥ mx_chk_loc_att
            # We have recorded the attractor with sufficient accuracy.
            # We now set the empty counters for the new attractor.
            cleanup_visited_cells!(bsn_nfo)
            bsn_nfo.visited_cell_label += 2
            bsn_nfo.current_att_label += 2
            reset_basins_counters!(bsn_nfo)
            # We return the label corresponding to the *basin* of the attractor
            return ic_label + 1
        end
        return 0
    end
```

This means that, once we enter `:att_found`, we start labelling **all encountered cells** as "attractor cells". However, the counter is incremented only the second time we encounter the attractor cells. And each time we do not encounter an attractor cell, we reset the counter and start again from scratch.

For a long time now I thought that the mode of operation is a second option: "every subsequent visited cell is labelled an attractor cell, and each time the counter is incremented." That's not what we have.

But my argument is: shouldn't we do the second option...? The way things are right now _we are searching for recurrences twice_: We first search for recurrences once to trigger the `:att_found` phase. But then **we search for recurrences again** during the `:att_found` phase, as only when a _second_ recurrence occurs, we increment the counter.

So, isn't it better if we do the second option? This way, we only search for recurrences once. Additionally, this way, the user has much better clarity of what value to give to `mx_chk_fnd_att`. It is the only parameter that relates with recurrences in state space.

Now, if we do go with the second option, it is not clear to me what `mx_chk_loc_att` should do... My guess is that it simply counts how many steps to take after having found an attractor to label cells as attractor cells.

Anyways, I am wondering, why did we do things this way. We search for recurrences twice. But it doesn't "improve the accuracy" of the algorithm if you think about it, because `:att_found` is a state one can never escape from. 

I am not sure. I hope you know this better and remember things better so you can tell me why the code is the way 1 instead of the way 2...

Datseris:
Closes #102.

EDIT: Updated.

This PR brings many positive changes. It originally was just a simple "fix #102" but this requires much more work... Details:

- Change the recurrences algorithm to do what is described in #102. This makes the user experience simpler, as the impact of the keyword `att_loc...` is simpler and completely separated from recurrences.
- Fixes a mistake with the lost counter interrupting the main counter of the FSM.
- Fixes a bug in the animation function for making a video of attractor continuation.
- Rewrites the irregular grid tests. Now they are much smaller, and simpler, and more accurate.
- Ensures all tests pass with the latest changes, including the extensive test suite.
- Updates many docstrings to be clearer and more precise.
- Enables a new type of error handling. We can identify when two attractors intersect in the grid, and throw an error instead of letting the simulation run forever!

### DIFF
--- a/src/mapping/recurrences/attractor_mapping_recurrences.jl
+++ b/src/mapping/recurrences/attractor_mapping_recurrences.jl
@@ -55,13 +55,14 @@ want to search for attractors in a lower dimensional subspace.
 
 ### Finite state machine configuration
 
-* `mx_chk_fnd_att = 1000`: Number of consecutive visits to previously visited
-  unlabeled cells required before declaring we have converged to a new attractor.
+* `mx_chk_fnd_att = 100`: Number of consecutive visits to previously visited
+  unlabeled cells (i.e., recurrences) required before declaring we have converged to a new attractor.
   This number tunes the accuracy of converging to attractors and should generally be high
   (and even higher for chaotic systems).
-* `mx_chk_loc_att = mx_chk_fnd_att÷10`: Number of subsequent steps taken to locate accurately the new
+* `mx_chk_loc_att = 1000`: Number of subsequent steps taken to locate accurately the new
   attractor after the convergence phase is over. Once `mx_chk_loc_att` steps have been
   taken, the new attractor has been identified with sufficient accuracy and iteration stops.
+  This number can be very high without much impact to overall performance, as this phase.
 * `store_once_per_cell = true`: Control if multiple points in state space that belong to
   the same cell are stored or not in the attractor, when a new attractor is found.
   If `true`, each visited cell will only store a point once, which is desirable for fixed
--- a/src/mapping/recurrences/attractor_mapping_recurrences.jl
+++ b/src/mapping/recurrences/attractor_mapping_recurrences.jl
@@ -2,22 +2,23 @@
 # Type definition and documentation
 #####################################################################################
 """
-    AttractorsViaRecurrences(ds::DynamicalSystem, grid::Tuple; kwargs...)
+    AttractorsViaRecurrences(ds::DynamicalSystem, grid; kwargs...)
 
 Map initial conditions of `ds` to attractors by identifying attractors on the fly based on
-recurrences in the state space, as outlined by Datseris & Wagemakers [Datseris2022](@cite).
+recurrences in the state space, as outlined in [Datseris2022](@cite).
 
 `grid` is instructions for partitioning the state space into finite-sized cells
 so that a finite state machine can operate on top of it. Possibilities are:
 
 1. A tuple of sorted `AbstractRange`s for a regular grid.
   Example is `grid = (xg, yg)` where `xg = yg = range(-5, 5; length = 100)`
-  for a two-dimensional system
+  for a two-dimensional system.
 2. A tuple of sorted `AbstractVector`s for an irregular grid, for example
-  `grid = (xg, yg)` with `xg = vcat(range(-5, -2; length = 50), range(-2, 5; length = 50)),
+  `grid = (xg, yg)` with `xg = range(0, 10.0^(1/2); length = 200).^2,
   yg = range(-5, 5; length = 100)`.
 3. An instance of the special grid type
-[`SubdividedBasedGrid`](@ref), which can be created either manually or by using [`subdivision_based_grid`](@ref).
+  [`SubdividedBasedGrid`](@ref), which can be created either manually or by using
+  [`subdivision_based_grid`](@ref).
   This automatically analyzes and adapts grid discretization
   levels in accordance with state space flow speed in different regions.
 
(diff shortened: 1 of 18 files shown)


Datseris:
Wow, only 1 test has failed! That is amazing.

(well, 3 tests have failed, but 2 of them are irregular grid tests that check how many points were stored in the attractor. the tests themselves don't make much sense, but I've also changed the default storage amount here anyways)

Datseris:
Extensive tests bring more failures. However, I will first away to see if @awage agrees with this change and finds it a step forwards or not, before investing any time to fix tests.

awage:
I am digging into the test results. I will try to see what is wrong with the henon example.

Datseris:
Sure, but please don't push something on this branch as I am working on it. If you find the solution simply report it as a comment.

Datseris:
By the way, this change is great even for us, not just for users. This new error clause I've added is very helpful. It detects very well when the algorithm has problems, either due to slowing down (like the problems @KalelR reported), or because the cells are too large and a cell may contain two attractors. I've tested this with the fast-slow systems, as well as with the Lorenz84 where the attractors intersect. _This is very good!_

This is an incredible quality of life change! We should focus on fixing the tests soon so that I merge this and tag this, because I am giving a workshop about Attractors.jl in 10 days.

Datseris:
If the error occurs, tjhat's the error message we get:

```
 During the phase of locating a new attractor, found via sufficient recurrences,
we encountered a cell of a previously-found attractor. This means that two
attractors intersect in the grid, or that the precision with which we find
and store attractors is not high enough. Either decrease the grid spacing,
or increase `mx_chk_fnd_att` (or both).

Index of cell that this occured at: CartesianIndex(1, 1).
```

awage:
Nice! It is really helpful. 

I found the bug. Due to the new method the lines between 181 and 186 should be: 

```
            current_basin = bsn_nfo.current_att_label + 1
            cleanup_visited_cells!(bsn_nfo)
            bsn_nfo.visited_cell_label += 2
            bsn_nfo.current_att_label += 2
            reset_basins_counters!(bsn_nfo)
            # We return the label corresponding to the *basin* of the attractor
            return current_basin
```

If you return ic_label you may have unexpected behavior .

awage:
Seems that all test passing except for the irregular grids!

Datseris:
@awage check out lines 281-288 of the source code:
```julia

    if next_state != current_state
        # reset counter except in lost state (the counter freezes in this case)
        if current_state == :lost
            bsn_nfo.consecutive_lost = 1
        else
            bsn_nfo.consecutive_match = 1
        end
    end
```

We have the comment that we "freeze" the lost counter, but in fact, we do the opposite. The counter is reset!!! What are we supposed to do? Freeze it? in this case we need to delete the `bsn_nfo.consecutive_lost = 1` line!

awage:
The comment is misleading. When you switch state the `current_state` and `next_state` are different. The internal counter `consecutive_match` is set to zero. 

The exception is when you move **out** of the `:lost` state: you set `consecutive_lost` to one but you do not reset `consecutive_match`. The algorithm pick up where it left and resume its activity. 

I am not satisfied with this part of the code. But I do not find anything better for now. The comment should be updated.

awage:
Maybe the way is to treat the `:lost` state separately. I will draft a PR.

awage:
see #105

awage:
I have done minimal changes but it is more readable now! I think the behavior of the system is the same. 

I am thinking about including a test specifically for the :lost state. What do you think?

awage:
This is a simple test for the lost state, part of the attractor lies outside the defined grid. What is amazing is that the algorithm finds its way out and identifies the attractor: 

```
    henon_rule(x, p, n) = SVector{2}(1.0 - p[1]*x[1]^2 + x[2], p[2]*x[1])
    henon() = DeterministicIteratedMap(henon_rule, zeros(2), [1.4, 0.3])
    ds = henon()
    xg = yg = range(-1.0, 1.0; length=100)
    grid = (xg, yg)
    mapper = AttractorsViaRecurrences(ds, grid)
    @show mapper([0. ,0.])
```

Datseris:
@awage or @KalelR this PR is now finished. I've updated the top level comment of the PR with all the changes this PR does: https://github.com/JuliaDynamics/Attractors.jl/pull/103#issue-1931705212 .

Please have a read, and give a review. If you give an accepting review, I'll merge. 

(I can't merge without a passing review, I've updated the repository settings so that all PRs need a passing review. It's good to ensure I don't just casually push stuff to `main` that I may regret later.)

src/mapping/recurrences/attractor_mapping_recurrences.jl
```diff
@@ -82,36 +88,38 @@ want to search for attractors in a lower dimensional subspace.
...
+
+-  The trajectory hits `mx_chk_fnd_att` times in a row previously visited cells:
    it is considered that an attractor is found and is labelled with a new ID. Then,
-   iteration continues a bit more until we have identified the attractor with sufficient
-   accuracy, i.e., until `mx_chk_loc_att` cells with the new ID have been visited.
+   iteration continues for `mx_chk_loc_att` steps. Each cell visited in this period stores
+   the "attractor" information. Then iteration terminates and the initial condition is
+   numbered with the attractor's ID.
 -  The trajectory hits an already identified attractor `mx_chk_att` consecutive times:
    the initial condition is numbered with the attractor's ID.
```
awage, line 109:
```suggestion
   the initial condition is numbered with the attractor's basin ID.
```

awage:
All good for me. I haven't checked the irregular grid stuff yet. I will do it another day.  But the rest is good to merge!

PR #105, opened 2023-10-09
Small improvements on update_finite_state_machine

awage:


Datseris (approved):

src/mapping/recurrences/finite_state_machine.jl
```diff
@@ -269,24 +270,29 @@ function update_finite_state_machine!(bsn_nfo, ic_label)
         # hit an attractor box
         next_state = :att_hit
     elseif ic_label == -1
-        # out of the grid we do not reset the counter of other state
-        # since the trajectory can follow an attractor that spans outside the grid
```
Datseris, line 273: why was this comment simplified? This is the only point in the source code that we say that we don't reset the other counters when lost. Please return this comment to its original state and then go ahead and merge the brand! @awage

Datseris: merge the branch *

PR #107, opened 2023-10-11
Rename some keywords of `AttractorsViaRecurrences`

### LINKED ISSUE #104: Rename keywords of `AttractorsViaRecurrences`
I believe the names of the keywrods of `AttractorsViaRecurrences` are not the best. The problems I notice are:

1. They use too many abbreviations
2. They are inconsistent: `mx_chk_hit_bas` vs `mx_chk_att`. Why one has hit and the other not?
3. They are mis-named. The keywords don't quantify the "maximum" number of consecutive hits. They quantify the _required_ number of consecutive hits. 
4. They never mention "recurrence" even though that is the central concept.

I propose the following renames:

- `mx_chk_fnd_att -> consecutive_recurrences`
- `mx_chk_loc_att -> attractor_locate_steps`
- `mx_chk_att -> consecutive_attractor_steps`
- `mx_chk_hit_bas -> consecutive_basin_steps`
- `mx_chk_lost -> consecutive_lost_steps`
- `mx_chk_safety -> maximum_iterations`

If we agree with these, once we are done with PR #103 , I will open a new PR that does the rename and add deprecations in place.

Datseris:
Closes  #104

This PR does the renaming of #104 and adds changelog entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# v1.13
+
+- A renaming of some of the options (keyword arguments) of `AttractorsViaRecurrences` has been done in line with the clarity increase of the algorithm. The following renames are in place and currently deprecated:
+  - `mx_chk_fnd_att -> consecutive_recurrences`
+  - `mx_chk_loc_att -> attractor_locate_steps`
+  - `mx_chk_att -> consecutive_attractor_steps`
+  - `mx_chk_hit_bas -> consecutive_basin_steps`
+  - `mx_chk_lost -> consecutive_lost_steps`
+  - `mx_chk_safety -> maximum_iterations`
+
+
 # v1.12
 - New algorithm `GroupViaPairwiseComparison` to group features in `AttractorsViaFeaturizing`. Simpler, typically faster and using less memory than DBSCAN, it can be useful in well-behaved systems.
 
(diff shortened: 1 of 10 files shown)


Datseris:
Yeah this system diverges to infinity:
```
julia> trajectory(ds, 100, [2.0])[1]
1-dimensional StateSpaceSet{Float64} with 101 points
       2.0
      -4.0
      56.0
 -175504.0
       5.405813486073056e15
      -1.5797311146339633e47
       3.942298606396109e141
     -Inf
     -Inf
     -Inf
       ⋮
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
     -Inf
```
Which makes sense of course, as x^3 is larger than x when x > 1...............

awage:
I have no problem with the keywords changes. But  is there a string doc in the code where we say that these keywords are equivalent to the old ones? It can be also a legacy `basins_of_attraction` function.

Datseris:
No, this is said in the CHANGELOG. Why would be in a docstring? The docstrings are only meaningful for the current version of the code. They have no reason to refer to prior versions.

Datseris:
@awage is it okay that this change is mentioned in the changelog?

awage:
Ok, I understand. So it is deprecated but the keywords are still accepted until the next breaking release? In the changelog it should be clear that the keywords are still accepted until the next breaking version. Then we can forget about them.

Datseris:
This is what "deprecated" means. Quite literally: "the names are still accepted until the next breaking release. They will be removed at the next breaking release. They are not mentioned anywhere in the documentation."

theogf:
I might be wrong but it looks like the old kwargs are not accepted anymore?

Datseris:
There should be. If not, it's a bug. Do you have a MWE? If yes, open an new issue!

PR #108, opened 2023-10-11
fix `AttractorsViaProximity` with `Inf` trajectories

Datseris:
- Pre-emtivively check for Inf
- Use `always_false` from nearest neighbors for better performance


Datseris:
@reykboerner you can give an approving review to this PR so we can merge it.

reykboerner (approved):
I've checked these changes and the test "Fix #61" now returns `-1` as desired. I cannot comment on the use of `always_false` from `NearestNeighbors` for performance but have verified that `mapper` works as before with this change.

PR #110, opened 2023-10-15
improve matching description

Datseris:
I was a bit unhappy that `RAFM` docstring didn't really have any description for matching. And the other two related docstrings were very long. This PR:

- Adds a small summary of matching in `RAFM` docstring.
- Streamlines the other two docstrings
- Adds the missing `use_vanished` keyword in `RAFM`.

### DIFF
--- a/src/continuation/match_attractor_ids.jl
+++ b/src/continuation/match_attractor_ids.jl
@@ -24,19 +24,9 @@ the dictionaries, by calling the [`replacement_map`](@ref) function directly.
 
 ## Description
 
-When finding attractors and their fractions in Attractors.jl,
-different attractors get assigned different IDs. However
-which attractor gets which ID is somewhat arbitrary. Finding the attractors of the
-same system for slightly different parameters could label "similar" attractors (at
-the different parameters) with different IDs.
-`match_statespacesets!` tries to "match" them by modifying the IDs,
-i.e., the keys of the given dictionaries. Do note however that there is nothing
-in this function that is limited to attractors in the formal mathematical sense.
-Any dictionary with `StateSpaceSet` values is a valid input and these sets
-may represent attractors, trajectories, group of features, or anything else.
-
-The matching happens according to the output of the [`setsofsets_distances`](@ref)
-function with the keyword `distance`. `distance` can be whatever that function accepts,
+The distance between all possible pairs of sets between the "old" and "new" containers
+is computed using [`setsofsets_distances`](@ref) with the keyword `distance`.
+`distance` can be whatever that function accepts,
 i.e., one of `Centroid, Hausdorff, StrictlyMinimumDistance` or any arbitrary user-
 provided function that given two sets it returns a positive number (their distance).
 State space sets are then matched according to this distance.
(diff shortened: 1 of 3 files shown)


src/continuation/match_attractor_ids.jl
```diff
@@ -24,19 +24,9 @@ the dictionaries, by calling the [`replacement_map`](@ref) function directly.
...
-may represent attractors, trajectories, group of features, or anything else.
-
-The matching happens according to the output of the [`setsofsets_distances`](@ref)
-function with the keyword `distance`. `distance` can be whatever that function accepts,
+The distance between all possible pairs of sets between the "old" and "new" containers
+is computed using [`setsofsets_distances`](@ref) with the keyword `distance`.
+`distance` can be whatever that function accepts,
 i.e., one of `Centroid, Hausdorff, StrictlyMinimumDistance` or any arbitrary user-
 provided function that given two sets it returns a positive number (their distance).
 State space sets are then matched according to this distance.
```
awage, line 32:
```suggestion
i.e., one of `Centroid, Hausdorff, StrictlyMinimumDistance` or any arbitrary user-
provided function that given two sets returns a positive number (their distance).
State space sets are then matched according to this distance. Notice that the 
algorithm do not check if the custom function is valid. It is left to the user to 
ensure that the function is correct.  
```

awage (approved):

src/continuation/match_attractor_ids.jl
Datseris: It is not clear to me what you mean by "valid" here? That the function doesn't error, or that the function doesn't return real numbers, or that it may return negative numbers? But in any case, I am not sure whether this is worth mentioning. The user provides the function so the user knows what the function does. Generic function input always works like that. E.g., the description for the function `findall` doesn't mention "we don't check your function".

awage: It is just a warning to avoid negative distance. Also the function should symmetric `f(a,b) = f(b,a)`. But if it is not common practice leave out this comment. There is a typo that needs to be changed: 

 that given two **sets returns** a positive number (their distance).

Datseris: The condition of symmetry does not have to be fullfilled, the matching algorithm does not assume or utilize this property. (and in fact, the positive definite condition also doesn't need to be fullfilled, internally we match from smallest to largest distance, and a negative distance would be the smallest). But yeah, it is not standard practice to impose instructions on the function in docstrings. Only the return type. That's the only thing Julia can know of anyways, the language cannot access the function internal operation when function is given as an input.

For the typo: is the typo the missing "it"? I think it is correct to say "given two sets _it_ returns the..." because "it" refers to the function.

awage: The subject is the function:  "any arbitrary user-provided function that  returns a positive number given two sets."

Datseris: ah yes you are right, I'll change it!

Datseris: Well, no, the sentence is correct:
"any arbitrary user-provided function that given two sets it returns a positive number (their distance)." I don't see anything wrong. It is better to have the positive number at the end because we can then explain that this number is the distance between the sets.

PR #120, opened 2024-01-16
Basins heatmap with a shading proportional to a matrix

awage:
This PR brings a new function to produce shading effects in basins. 

It has a custom colormap that goes from light to dark colors according to an iteration matrix. The iterations of the mapper are stored with the new `get_iterations` function. It uses some internal variables of the mapper. 

The effect are these gorgeous figures: 

Captura de pantalla_2024-01-16_15-00-52
Captura de pantalla_2024-01-16_15-01-57

Test script: 
```julia
using CairoMakie
using Attractors, StaticArrays
function henon_rule(u, p, n) 
    x, y = u
    a, b = p
    xn = 1.0 - a*x^2 + y
    yn = b*x
    return SVector(xn, yn)
end
u0 = [0.2, 0.3];
p0 = [1.2, 0.3];
henon = DeterministicIteratedMap(henon_rule, u0, p0);
xg = range(-2.5, 2.5; length = 1500)
yg = range(-2.5, 2.5; length = 1500)
grid=(xg,yg)
mapper = AttractorsViaRecurrences(henon, grid; sparse = false)
# basins, attractors = basins_of_attraction(mapper; show_progress = false)
basins = zeros(Int8, 1500, 1500); iterations = zeros(Int32,1500,1500)
for (i,x) in enumerate(xg), (j,y) in enumerate(yg) 
    l = mapper([x,y])
    n = get_iterations(mapper)
    basins[i,j] = l
    iterations[i,j] = n
end
using StatsBase:median
# Display basins with shading. 
shaded_basins_heatmap(grid, basins, iterations, attractors;  show_attractors = false, maxit = 2*median(iterations), title = "test" )
```


awage:
I left some comments in the code on how the gradient is constructed: 

```julia
    # The colormap is constructed in such a way that the first color maps 
    # from id to id+0.99, id is an integer describing the current basin.
    # Each id has a specific color associated and the gradient goes from 
    # light color (value id) to dark color (value id+0.99). It produces
    # a shading proportional to a value associated to a specific pixel. 

```

There is some code redundancy between `heatmap_attractors` and this plotting method. I don't know if it is worth factorizing the two functions. 

The colors are also differents. I used named colors from Colors.jl. It is more convenient.

ext/plotting.jl
```diff
@@ -85,6 +85,121 @@ function Attractors.heatmap_basins_attractors!(ax, grid, basins, attractors;
...
+        :lavenderblush,
+        :lightyellow,]      
+    DARK_COLORS = [
+        :black,
+        :red4,
+        :darkgreen,
+        :navyblue,
+        :purple4,
+        :gold4,]      
+    n = length(LIGHT_COLORS)
```
Datseris, line 124: This would make the same colors if there are more than 6 attractors, right?

ext/plotting.jl
```diff
@@ -85,6 +85,121 @@ function Attractors.heatmap_basins_attractors!(ax, grid, basins, attractors;
...
+function colors_from_keys_shaded(ukeys)
+    # Unfortunately, until `to_color` works with `Cycled`,
+    # we need to explicitly add here some default colors...
+    COLORS =[ 
+        :black,
+        :red,
+        :green, 
+        :blue,
+        :purple,
+        :yellow,]
```
Datseris, line 97: Let's use the colors we already use please. There is a simple function that given `ukeys` it gives you colors. If you want to make these colors lighter or darker, use the function:
```
"""
    darken_color(c, f = 1.2)
Darken given color `c` by a factor `f`.
If `f` is less than 1, the color is lightened instead.
"""
function darken_color(c, f = 1.2)
    c = to_color(c)
    return RGBAf(clamp.((c.r/f, c.g/f, c.b/f, c.alpha), 0, 1)...)
end
```

This way, we also resolve the problem that we only have 6 colors, as the function I mention works for any number of attractors.

src/plotting.jl
```diff
@@ -15,6 +15,28 @@ function heatmap_basins_attractors end
 function heatmap_basins_attractors! end
 export heatmap_basins_attractors, heatmap_basins_attractors!
 
+""" 
+    shaded_basins_heatmap(grid, basins, iterations, attractors; kwargs...) 
+
```
Datseris, line 20: This is a specialized application. It really only makes sense for 2D systems (or systems projected into 2D). 

That is why I think you should take the following code:
```
# basins, attractors = basins_of_attraction(mapper; show_progress = false)
basins = zeros(Int8, 1500, 1500); iterations = zeros(Int32,1500,1500)
for (i,x) in enumerate(xg), (j,y) in enumerate(yg) 
    l = mapper([x,y])
    n = get_iterations(mapper)
    basins[i,j] = l
    iterations[i,j] = n
end
using StatsBase:median
```

and make it into a function `iterates_of_basins_of_attraction` that takes the mapper and grid and returns `basins, attractors, iterations`. This way you can give the output of the fnction directly into the plotted function.

Datseris (commented):
This is really awesome! But I think we can make the functionality even more convenient!

Additionally, with my suggestion, the function `get_iterations` remains internal, it is not exposed to the user.

Lastly, `get_iterations` is a very generic name. Let's rename it to `iterations_to_converge` please!

ext/plotting.jl
awage: The colors are circled mod 6. So no problem on this side. Anyway I will stick to the defined colors and adjust the brightness for each one.

awage:
Classic DS color scheme: 
Captura de pantalla_2024-01-22_22-09-43

awage:
Works for more attractors: 
Captura de pantalla_2024-01-22_22-23-53

awage:
Do we need to include the test example? Or just for the docs? 

```
using CairoMakie
using Attractors
using PredefinedDynamicalSystems

ds = Systems.henon([0.2, 0.3]; a = 1.2, b = 0.3)
xg = yg =  range(-2.5, 2.5; length = 1500)
grid=(xg,yg)
mapper = AttractorsViaRecurrences(ds, grid; sparse = false)
basins,attractors, iterations = iterates_of_basins_of_attraction(mapper, grid)

using StatsBase:median
# Display basins with shading. 
shaded_basins_heatmap(grid, basins, iterations, attractors;  show_attractors = true, maxit = 2*median(iterations), title = "test" )
```

awage:
Almost finished. The function `iterates_of_basins_of_attraction` should work on any grid. But the plot is only possible in 2D.

Datseris:
@awage thanks, I just re-arranged the docs. Since this PR adds and exports a new public function `iterates_of_basins_of_attraction`, this function also needs to be tested in the test suite. Would you mind writting a quick test please?

Datseris:
Hm, even though we fixed documentation build in a different PR, in this one it fails, perhaps you need to update this PR with current `main`?

awage:
Ok, I will merge main into this branch.

Datseris:
@awage can you please add the line `using OrdinaryDiffEq: Vern9` in the file `src/examples.md:882-893` just before the line `function fitzhugh_nagumo(u,p,t)` ? this is causing the doc failutre (which is unrteated with your PR, it was a mistake from the PR #61 )

awage:
Done! Lets see if it pass the tests.

docs/make.jl
```diff
@@ -3,6 +3,7 @@ using Attractors
 using Attractors.DynamicalSystemsBase
 using Attractors.StateSpaceSets
 using PredefinedDynamicalSystems
+using OrdinaryDiffEq:Vern9
```
Datseris, line 6: This does not need to be in the make.jl file. For dependencies used within the `@example` blocks in the docs, add a dependency in `docs/Project.toml`.

Datseris:
I'll fix and merge this now quickly

Datseris (approved):

PR #125, opened 2024-04-29
Improvements to convergence time functions.

Datseris:
Add new function `convergence_and_basins_fractions`, and improve the algorithm
counting the convergence iterations for `AttractorsViaRecurrences` to give
more accurate results, by taking into account the user-provided convergence criteria.

To make it possible i added one more field on the basin info struct that contains a return code that tells us how the recurrences map ended.


src/mapping/recurrences/attractor_mapping_recurrences.jl
```diff
@@ -174,7 +174,20 @@ end
...
+    i = m.bsn_nfo.safety_counter
+    kw = m.kwargs
+    if m.bsn_nfo.return_code == :new_att
+        # in this scenario we have an addition amount of iterations that is
+        # equal to the recurrences steps. We subtract this for more correct
+        # estimation of convergence time that will be closer to the convergence
+        # time of neighboring grid cells.
+        x = get(kw, :consecutive_recurrences, 100) + get(kw, :attractor_locate_steps, 1000)
+    else
+        x = get(kw, :consecutive_attractor_steps, 2)
```
awage, line 187: elseif m.bsn_nfo.return_code == :bas_hit
   x = 0

Datseris: I think we should actually return `NaN` in the `:bas_hit` case. because in this scenario the reported convergence time can be wildly inaccurate. We really can't even quantify how inaccurate it will be.!

Datseris:
If you wish to contribute a test that indeed would be fantastic!

Datseris:
I also realized that in the case of the Proximity Mapper, we return the actual convergence time, not iteration number.
While in the recurrences we return the iteration number instead. We have to harmonize this. 

First, we need to rename this function to `convergence_time` and second, for the recurrences we need to return the iteration multiplied by the internal; `Δt`. I will do this change now.

PR #126, opened 2024-05-30
(2.0) Parameter-by-parameter continuation using featurizing 

KalelR:
Hey, this is a new PR, following up on #69. Sorry I took so long, made all of this a bit unnecessarily complicated.  I came back to this now as I need this continuation for two projects I'm working on. On the previous PR, we were discussing whether it was better to match the attractors found for different parameters using the clouds of features or the attractors themselves. There were good points for using the features, one of them being that there would be no need for reconstructing the attractors. This is true for matching, but, importantly, not for seeding. We need the attractors to sample ics from them to use them as seeds for the attractor-finding algorithm in the subsequent parameter. Thus, since we need the attractors anyway for seeding (which is the heart of the continuation!), then we might as well use them for matching! If the user then wants to match via features, they can just extract the features in their matching function.

In summary, this PR thus implements continuation for AttractorsViaFeaturizing very similarly to how it is implemented for AttractorsViaRecurrences. The only big difference is that the `ics` sent to `continuation` must be pre-generated and stored in a StateSpaceSet, contrary to the sampler function that the AttractorsViaRecurrences version needs. This is because `basins_fractions` requires the `ics` in this way to be able to extract the attractors after grouping the features. Again, the attractors are needed for seeding, so this is necessary.
After attractors are found, initial conditions are sampled from them and sent together with the pre-generated `ics` for `basins_fractions` on the next parameter. The ics are all put together for extracting features and grouping them. 

Let me know what you think! 

I compared this algorithm with the recurrences one for the magnetic pendulum example in the documentations. Here is a comparison
image
with the recurrences on the left and the featurizing on the right for 1000 ics per parameter. I think the curves for featurizing are smoother because the `pre-generated` ics are fixed. For recurrences, new `ics` are generated for all parameters, which probably causes the bigger fluctuations.

### DIFF
--- a/src/continuation/continuation_featurizing_find_and_match.jl
+++ b/src/continuation/continuation_featurizing_find_and_match.jl
@@ -1,7 +1,64 @@
-export FeaturizingFindAndMatch
+export FeaturizingFindAndMatch, FFAM
 import ProgressMeter
 using Random: MersenneTwister
 
+"""
+    FeaturizingFindAndMatch <: AttractorsBasinsContinuation
+    FeaturizingFindAndMatch(mapper::AttractorsViaFeaturizing; kwargs...)
+
+A method for [`continuation`](@ref) as in [Datseris2023](@cite) that uses the featurizing algorithm for finding attractors ([`AttractorsViaFeaturizing`](@ref))
+and the "matching attractors" functionality offered by [`match_continuation!`](@ref). Based heavily on the `RecurrencesFindAndMatch`, which uses the recurrences algorithm ([`AttractorsViaRecurrences`](@ref)).
+
+You can use `FFAM` as an alias.
+
+## Keyword arguments
+
+- `distance = Centroid(), threshold = Inf, use_vanished = !isinf(threshold)`:
+  propagated to [`match_continuation!`](@ref).
+- `info_extraction = identity`: A function that takes as an input an attractor (`StateSpaceSet`)
+  and outputs whatever information should be stored. It is used to return the
+  `attractors_info` in [`continuation`](@ref). Note that the same attractors that
+  are stored in `attractors_info` are also used to perform the matching in
+  [`match_continuation!`](@ref), hence this keyword should be altered with care.
+- `seeds_from_attractor`: A function that takes as an input an attractor and the number of
+  initial conditions to sample from the attractor, and returns a vector with the sampled initial
+  conditions. By default, we just select the initial conditions randomly from the attractor.
+
+## Description
+
+At the first parameter slice of the continuation process, attractors and their fractions
+are found as described in the [`AttractorsViaFeaturizing`](@ref) mapper identifying groups
+of features. At each subsequent parameter slice, initial conditions are seeded from the
+ previously found attractors. They are put together with the pre-generated `ics` given to
+ `continuation`. Then, all of them are passed onto [`basins_fractions`](@ref) for
+ [`AttractorsViaFeaturizing`](@ref), which featurizes and groups them to identify the
+ attractors. 
+
+Naturally, during this step new attractors may be found. Once the basins fractions are
+computed, the parameter is incremented again and we perform the steps as before.
+
+This process continues for all parameter values. After all parameters are exhausted,
+the found attractors (and their fractions) are "matched" to the previous ones.
+I.e., their _IDs are changed_, so that attractors that are "similar" to those at a
+previous parameter get assigned the same ID.
+Matching is done by the [`match_continuation!`](@ref) function and is an _orthogonal_
+step. This means, that if you don't like the initial
+outcome of the matching process, you may call [`match_continuation!`](@ref) again
+on the outcome with different matching-related keywords.
+You do not need to compute attractors and basins again!
+
+Matching is a very sophisticated process that can be understood in detail by reading
+the docstrings of [`match_statespacesets!`](@ref) first and then [`match_continuation!`](@ref).
+Here is a short summary: attractors from previous and current parameter are matched
+based on their "distance". By default this is distance in state space, but any measure of
+"distance" may be used, such as the distance between Lyapunov spectra.
+Matching prioritizes new->old pairs with smallest distance: once these are matched
+the next available new->old pair with smallest distance is matched, until all new/old
+attractors have been matched. The `threshold` keyword establishes that attractors with
+distance > `threshold` do not get matched.
+Additionally, use `use_vanished = true` if you want to include as matching candidates
+attractors that have vanished during the continuation process.
+"""
 struct FeaturizingFindAndMatch{A, M, R<:Real, S, E} <: AttractorsBasinsContinuation
     mapper::A
     distance::M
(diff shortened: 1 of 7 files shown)


Datseris:
thanks, i will review this **after** Tuesday.

src/mapping/grouping/attractor_mapping_featurizing.jl
```diff
@@ -95,18 +95,23 @@ ValidICS = Union{AbstractStateSpaceSet, Function}
...
-function basins_fractions(mapper::AttractorsViaFeaturizing, ics::ValidICS;
-        show_progress = true, N = 1000
-    )
+function basins_fractions(mapper::AttractorsViaFeaturizing, og_ics::ValidICS;
+    show_progress = true, N = 1000, additional_ics::Union{ValidICS, Nothing} = nothing,
+)
+    if typeof(additional_ics) <: ValidICS
+        all_ics = deepcopy(og_ics)
+        append!(all_ics, additional_ics)
+        ics = all_ics
```
Datseris, line 104: How would this code work if `og_ics` is a function that spits out initial conditions...? `ValidICS` is either a function or a statespaceset.

Datseris:
Can you point me to where in the source code you "find the attractor" or "extract the attractor"?

src/mapping/grouping/attractor_mapping_featurizing.jl
KalelR: true, this is only useful for pre-generated `ics`. The line should be 
```julia
if typeof(additional_ics) <: StateSpaceSet
```

KalelR:
> Can you point me to where in the source code you "find the attractor" or "extract the attractor"?

The attractors are extracted as usual in `extract_attractors`, defined in https://github.com/KalelR/Attractors.jl/blob/28f662a29972c3b826bbaaa63b2cc44efce4e97f/src/mapping/grouping/attractor_mapping_featurizing.jl#L180.

Datseris:
I am kinda confused by how the matching works here... "what" are you matching on? I see there is an `info_extraction` function that extracts the attractors. But if you are matching on the attractors, there shouldn't be any influence from the user on that.

src/continuation/continuation_featurizing_find_and_match.jl
```diff
@@ -0,0 +1,148 @@
...
+        fam::FeaturizingFindAndMatch,
+        prange, pidx, ics;
+        samples_per_parameter = 100, show_progress = true, keep_track_maximum=true,
+    )
+    progress = ProgressMeter.Progress(length(prange);
+        desc="Continuating basins fractions:", enabled=show_progress
+    )
+
+    if ics isa Function
+        error("`ics` needs to be a Dataset.")
```
Datseris, line 98:
```suggestion
        throw(ArgumenteError("`ics` needs to be a StateSpaceSet."))
```

src/continuation/continuation_featurizing_find_and_match.jl
```diff
@@ -0,0 +1,148 @@
...
+  [`match_continuation!`](@ref), hence this keyword should be altered with care.
+- `seeds_from_attractor`: A function that takes as an input an attractor and the number of
+  initial conditions to sample from the attractor, and returns a vector with the sampled initial
+  conditions. By default, we just select the initial conditions randomly from the attractor.
+
+## Description
+
+At the first parameter slice of the continuation process, attractors and their fractions
+are found as described in the [`AttractorsViaFeaturizing`](@ref) mapper identifying groups
+of features. At each subsequent parameter slice, initial conditions are seeded from the
```
Datseris, line 31:
```suggestion
of features. From each group of features, an attractor is extracted by sampling
one point in the group, evolving it forwards in time, and collecting the trajectory.
That is why `ics` in `continuation` in this method must be a fixed set, and
not a sampling function.

Then, at each subsequent parameter slice, initial conditions are seeded from the
```

src/continuation/continuation_featurizing_find_and_match.jl
```diff
@@ -0,0 +1,148 @@
...
+ previously found attractors. They are put together with the pre-generated `ics` given to
```
Datseris, line 32:
```suggestion
 previously found attractors. They are put together with the provided `ics` given to
```

src/continuation/continuation_featurizing_find_and_match.jl
```diff
@@ -0,0 +1,148 @@
...
+the docstrings of [`match_statespacesets!`](@ref) first and then [`match_continuation!`](@ref).
+Here is a short summary: attractors from previous and current parameter are matched
+based on their "distance". By default this is distance in state space, but any measure of
+"distance" may be used, such as the distance between Lyapunov spectra.
+Matching prioritizes new->old pairs with smallest distance: once these are matched
+the next available new->old pair with smallest distance is matched, until all new/old
+attractors have been matched. The `threshold` keyword establishes that attractors with
+distance > `threshold` do not get matched.
+Additionally, use `use_vanished = true` if you want to include as matching candidates
+attractors that have vanished during the continuation process.
```
Datseris, line 60: Let's not duplicate this. Remove this whole copy pasted text in favor of just "matching works exactly as in [`RecurrencesFindAndMatch`] so see that docstring for more."

Datseris (approved):
Sure, seems fine with me. I have some comments on the docs.

awage:
Hello, I'm free from academic duties! I will review the code tomorrow.

src/continuation/continuation_featurizing_find_and_match.jl
```diff
@@ -0,0 +1,134 @@
...
+- `seeds_from_attractor`: A function that takes as an input an attractor and the number of
+  initial conditions to sample from the attractor, and returns a vector with the sampled initial
+  conditions. By default, we just select the initial conditions randomly from the attractor.
+
+## Description
+
+At the first parameter slice of the continuation process, attractors and their fractions
+are found as described in the [`AttractorsViaFeaturizing`](@ref) mapper identifying groups
+of features. From each group of features, an attractor is extracted by sampling one point
+in the group, evolving it forwards in time, and collecting the trajectory. That is why
```
awage, line 32: What is the sampling exactly? Is it one of the initial conditions belonging to a particular feature group?

awage:
Hi Kalel! I am not sure to get the point of sampling the attractors point. The thing that puzzles me is that you add these extra `ics` to the pool of predefined initial conditions. How this influences the mapper? Is it to ensure that you have at least one initial condition from each attractor? Because the mapper does its job with or without these extra initial conditions.

The matching is done the usual way. So what is the advantage over the previous matching continuation? 

The code is fine. It's just that I don't see the benefits right now over the previous method.

Datseris:
> How this influences the mapper? Is it to ensure that you have at least one initial condition from each attractor? Because the mapper does its job with or without these extra initial conditions.

It is the same as the **seeding** part of our original Recurrences continuation, the one we described in our 2023 paper. 

The featurizing mapper did not so far seed initial conditions from found attractors when moving to the next parameter. With this PR it does. This makes it more likely that if an attractor continues to exist, it will be found during the continuation.

awage:
Ok, I see more clearly the context. The code is good to go! 

@KalelR  @Datseris: I have an idea to factorize a bit the code. But it is a breaking change. A new type called `AttractorsFindAndMatch` can generalize the interface. Then we dispatch on the type of mapper depending if its recurences or featurizing. There some code in common that can be refactored. What do you think?

Datseris:
@awage have you seen #130  ...? I do exactly that. In fact, the whole code is already implemented and pushed in the branch `matcher_interface`. I will open a PR now.

awage:
> @awage have you seen #130 ...? I do exactly that. In fact, the whole code is already implemented and pushed in the branch `matcher_interface`. I will open a PR now.

Wow! I l'll  catch up.

Datseris:
See #132

KalelR:
Now with PR #132 and the correction at PR #146 this code is obsolete, correct? Shall we close this and PR #69 ?

Datseris:
Yeap.

PR #128, opened 2024-06-14
Example for enriching continuation with arbitrary other library functions

Datseris:
Alright, this is really the cream of the crop. This is really what DynamicalSystems.jl is about.

This PR introduces a new example that shows how one can enrich our continuation output with arbitrary functions of the library. Here lyapunov exponents and minimal fatal shock.

To make this more streamlined, I added a new plotting function as well as exported the function that resets the recurrences mapper.

This new example is what I would recommend using to show the power of the library.


Datseris:
closing in favor of #129

Issue #130, opened 2024-06-25
Interface redesign proposal: matchers and a generic continuation type


Datseris:
This proposal is motivated by two things:

1. The currently open PR #126 
2. Some offline discussion I had with Kalel about a new way to mattch attractors, that is fundamentally different from the way we match currently. It is so fundamentally different, that one cannot just adjust the "distance function" in our current `match_statespacesets!` function so that the same result is achieved.

I want to propose a new interface, that will naturally make Attractors.jl easier to extend both in continuation algorithms, and in matching algorithms. 

## Matchers

The first step is a complete separation of continuation and matching. In fact, this is already achieved in the source code, but not at a formal interface level.

I propose a new abstract type `SSSetMatcher` (state space set, a.k.a. attractor, matcher). Subtypes of this interface extend the function
```julia
function match_continuation!(attractors_info, matcher::SSSetMatcher)
    # do your thing, store the replacement maps
    return rmaps::Vector{Dict}
end
```
This function operates given the "attractors" (vector of dictionaries). The extra convenience function 
```julia
function match_continuation!(continuation_quantity::Vector{Dict}, rmaps::Vector{Dict})
    for (quantity, rmap) in zip(continuation_quantity, rmaps)
        swap_dict_keys!(quantity, rmap)
    end
    return
end
```
can be applied instantly to the basins fractions curves, or any other measure of attractors that we track, such as those in the example of PR #129 .

This new interface allows us to have the current matcher as one type, and the new matcher that Kalel has designed (but is not yet pushed anywhere online here) as a second matcher.

## New generic continuation type

This orthogonality of matchers and mappers means that we can generalize the current `RecurrencesFindAndMatch` to a more general `AttractorsContinueAndMatch`. An alternative name could be `AttractorsSeedAndMatch` or `AttractorsSeedContinueMatch`. let me know which one you prefer. 

Here is its docstring:

> 
>     AttractorsContinueAndMatch(mapper, matcher [, seeding])
> 
> A continuation method for [`continuation`](@ref).
> `mapper` is any subtype of [`AttractorMapper`](@ref) which implements
> [`extract_attractors`](@ref), while `matcher` is any subtype of [`SSSetMatcher`](@ref).
> 
> ## Description
> 
> This global continuation method has a simple basis.
> At the first parameter slice of the continuation process, attractors and their fractions
> are found using the given `mapper`.
> At each subsequent parameter slice,
> new attractors are found by seeding initial conditions from the previously found
[...]

PR #132, opened 2024-06-27
New generic continuation type and new matcher interface

Datseris:
This PR implements the ideas described in #130. There are still some TODOs left.

UPDATE: the matcher infrasturcture is hidden. After talking with Kalel we realized that this new matcher is actually a whole new continuation and is very simple to implement. This PR has the starting code for it but doesn't export anything new yet.

So this PR now focuses on generalizing the seed and match continuation, as well as renames and re-organization, see CHANGELOG!

This PR also removes the "info extraction" business from the seed and match continuation. This appears completely obsolete, as you can always extract any info you want from the attractors after the continuation, with the same cost as doing it during the continuation.

This PR also further improves the matching code to not unecessarily increment IDs in the mathcing process. That means: if we had previous attractors 1, 2, and current attractors 2 and 3, and attractor 3 matches to attractor 2, then the new attractor 2 won't get ID 4, it will get ID 3 as this is the next available integer.

- [x] implement `retract_keys`
- [x] Update documentation `/docs` files
- [x] add tests for the new interface
- [x] ensure previous code works and is deprecated
- [x] global rename `fractions_curves` to `fractions_cont` (for continuation)
- [x] global rename `attractors_info` to `attractors_cont` (for continuation)


src/continuation/continuation_afam.jl
```diff
@@ -0,0 +1,107 @@
...
+In code, matching is a rather trivial call to [`match_continuation!`](@ref)
+at the end of the `continuation` function.
+If you don't like the final matching output,
+you may use arbitrarily different `matcher` in [`match_continuation!`](@ref)
+without having to recompute the whole continuation.
+
+The matching algorithms are rather sophisticated,
+so how matching works is described in the docstrings of each `matcher`.
+"""
+struct AttractorsContinueAndMatch{A, M, S} <: AttractorsBasinsContinuation
```
awage, line 47: I like `AttractorsContinueAndMatch` , `AttractorsSeedAndMatch` do not refer to the continuation or matching process. In this new interface the seeding can be a random state space point generator also.

src/continuation/continuation_afam.jl
```diff
@@ -0,0 +1,107 @@
+"""
+    AttractorsContinueAndMatch(mapper, matcher [, seeding])
```
awage, line 2: The `seeding` argument is optional? In the type implementation it seems mandatory to declare seeding

src/continuation/continuation_afam.jl
```diff
@@ -0,0 +1,107 @@
...
+        )
+        # We do not match attractors here; the matching is independent step done at the end
+        current_attractors = deepcopy(extract_attractors(mapper))
+        push!(fractions_curves, fs)
+        push!(attractors_info, current_attractors)
+        overwrite_dict!(prev_attractors, current_attractors)
+        ProgressMeter.next!(progress; showvalues = [("previous parameter", p),])
+    end
+    # Match attractors (and basins)
+    rmaps = match_continuation!(attractors, matcher)
```
awage, line 104:
```suggestion
    rmaps = match_continuation!(attractors, acam.matcher)
```

src/matching/matching_interface.jl
```diff
@@ -0,0 +1,158 @@
+"""
+    SSSetMatcher
```
awage, line 2: I would simplified the name `SSSetMatcher` to `SetMatcher`. Unless you have a good reason to add State Space, it is easier to remember.

src/matching/sssdistance.jl
```diff
@@ -0,0 +1,100 @@
...
+is larger than this `threshold`, then it is guaranteed that the sets will get assigned
+different ID in the replacement map (the next available integer).
+"""
+@kwdef struct MatchBySSDistance{M, T<:Real}
+    distance::M = Centroid()
+    threshold::T = Inf
+end
+
+function replacement_map(a₊::AbstractDict, a₋, matcher::SSSetMatcher; next_id = max(maximum(keys₊), maximum(keys₋)) + 1, pi = nothing)
+    distances = setsofsets_distances(a₊, a₋, distance)
```
awage, line 49:
```suggestion
    distances = setsofsets_distances(a₊, a₋, matcher.distance)
```

src/continuation/continuation_afam.jl
```diff
@@ -0,0 +1,107 @@
...
+    match_continuation!(rmaps, fractions_curves)
```
awage, line 105: This function should be a low level function such as `replace_key_dict` or something similar.

src/continuation/continuation_afam.jl
awage: OK, I got it. The concrete implementations are still to be implemented. There will be several implementations depending on the type of mapper and matcher.

awage:
The general architecture is good. I left some minor comments. If you need further help please ask.

src/continuation/continuation_afam.jl
Datseris: I actually really want "seeding" to be part of the name because really that is what is unique about this continuation.

src/matching/match_basins.jl
```diff
@@ -0,0 +1,41 @@
+# TODO: I don't know how this fits into the new interface...
```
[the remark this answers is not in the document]
Datseris: Can you show me how it can be adapted to the new interface with a concrete example? In the current interface the inputs are dictionaries of IDs => values. Here we have an array whose entries are the IDs. Fundamentally different.

One idea I have for adapting to new interface is to transform the basin to dictionarys ID => vector of cartesian indices with this ID.

src/continuation/continuation_afam.jl
Datseris: Which function should be low level? `match_continuation!`?

awage: The second call: 

`match_continuation!(rmaps, fractions_curves)`

It is just bookkeeping

src/matching/match_basins.jl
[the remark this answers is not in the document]
awage: Ok, tomorrow I will work on this one.

Datseris:
alright, this is in a very good state now. the "retract keys" functionality is missing if someone can port it over!

src/continuation/continuation_afam.jl
Datseris: I have renamed the function to `match_sequentially!`, is this okay?

src/matching/match_basins.jl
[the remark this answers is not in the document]
Datseris: @awage haha have a look at the latest commit I just pushed. I was working on this today and I ended up doing **exactly** the same thing you thought. :D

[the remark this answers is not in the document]
Datseris: This method is actually just as efficient as the previous method. Why do you think it is less efficient?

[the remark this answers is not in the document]
awage: Not efficient in terms of extra memory allocation. But this is not the computational bottleneck. Basins computation is.

src/matching/match_basins.jl
```diff
@@ -0,0 +1,81 @@
...
+    # propagates this "distance" to the matching code of `MatchBySSDistance`!
+    keys₊, keys₋ = keys.((a₊, a₋))
+    nextid = isnothing(next_id) ? max(maximum(keys₊), maximum(keys₋)) + 1 : next_id
+
+    distances = Dict{eltype(keys₊), Dict{eltype(keys₋), Float64}}()
+    for i in keys₊
+        Bi = a₊[i]
+        d = valtype(distances)() # d is a dictionary of distances
+        # Compute normalized overlaps of each basin with each other basis
+        for j in ids₋
```
awage, line 65:
```suggestion
        for j in keys₋
```

src/matching/match_basins.jl
```diff
@@ -0,0 +1,81 @@
...
+            Bj = b₋[j]
```
awage, line 66:
```suggestion
            Bj = a₋[j]
```

src/matching/match_basins.jl
```diff
@@ -0,0 +1,81 @@
...
+representing basins of attraction after and before, and the conversion to dictionaries
+will happen internally as it is supposed to.
+To replace the `IDs` in `b₊` given the replacement map just call `replace!(b₊, rmap...)`,
+or use the in-place version [`replacement_map!`](@ref) directly.
+"""
+struct MatchByBasinOverlap
+    threshold::Float64 = Inf
+end
+
+function replacement_map(b₊::AbstractArray, b₋::AbstractArray, matcher::MatchByBasinOverlap; i = nothing)
```
awage, line 32:
```suggestion
function replacement_map(b₊::AbstractArray, b₋::AbstractArray, matcher = MatchByBasinOverlap(); i = nothing)
```
Maybe leaving a default structure makes the function more user friendly.

KalelR:
>     * [ ]  Implement the new matcher type "by basin enclosure"

I'm not sure if we need to implement a new matcher. In issue #133 I described how I'm doing this matching by basin enclosure. Although it is almost definitely not the best way of doing it, it has the benefit of just adding a new function, and not changing the continuation or matching functions.

src/continuation/continuation_afam.jl
awage: OK! 
```suggestion
    match_sequentially!(fractions_curves, rmaps)
```

src/matching/match_basins.jl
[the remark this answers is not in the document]
Datseris: There is no extra memory allocation happening. The operation `findall(isequal(k), b₋) for k in unique(b₋)` is always done in the end of the day regardless of how we will represent the basins.

Datseris:
I will wrap this up this weekend and tag a new release with the updated tutorial, if you have any crucial comments please add them here asap.

Datseris:
@awage @KalelR alright, I think this is some really good work on the generic global continuation method. Have a look at the documentation:

https://github.com/JuliaDynamics/Attractors.jl/blob/41a1feab09f93e0a8b65fa561318c7a8e8743edb/src/continuation/continuation_ascm_generic.jl#L5-L94

@KalelR this method is so general, that now we _truly_ don't need to create a new continuation type "basin enclosure". The only thing we have to do is to add a dispatch to the function:

https://github.com/JuliaDynamics/Attractors.jl/blob/matcher_interface/src/matching/matching_interface.jl#L29-L38

Do you think you can create this function from your original code and PR it? Everything else will just work and you only need to define this function to have basin enclosure. You can have it with or without seeding, and with any mapper, whether it is featurizing or recurrences. 

I improved the featurizing code to always store the attractors via evolving the trajectories, irrespectively if the sampler is random or pre-described initial condiitons. I also think that your preferred grouping method with "PairwiseComparison" will benefit from the seeding, as the **first initial conditions** will likely be the attractors, so they will form the center of the clusters for you.

src/matching/matching_interface.jl
```diff
@@ -0,0 +1,193 @@
...
+
+# For now, the only parts exposed are these functions:
+export replacement_map, replacement_map!, match_sequentially!
+# all of which take as input the treshold and distance of the
+# `MatchBySSSetDistance` matcher.
+
+"""
+    IDMatcher
+
+Supertype of all "matchers" that match can IDs labelling attractors.
```
awage, line 13:
```suggestion
Supertype of all "matchers" that match IDs to attractors.
```

src/matching/matching_interface.jl
```diff
@@ -0,0 +1,193 @@
...
+
+Matchers implement an extendable interface.
+For simple matcher, one only needs to extend the function [`replacement_map`](@ref).
+This function is used by the higher level function [`match_sequentially!`](@ref).
+Matchers that may have more involved behavior requiring the dynamical system and
+current and previous parameter need to extend the internal function
+[`_match_attractors`](@ref). This function is utilized by [`AttractorSeedContinueMatch`](@ref).
+
+As you user you typically only care about given an instance of `IDMatcher`
```
awage, line 22:
```suggestion
As a user you typically only care about given an instance of `IDMatcher`
```

src/matching/matching_interface.jl
```diff
@@ -0,0 +1,193 @@
...
+abstract type IDMatcher end
+
+# lowest level function API that falls back to `replacement_map`
+"""
+    _match_attractors(
+        current_attractors, prev_attractors, matcher,
+        ds, p, pprev
+    )
+
+Return the replacement map, mapping IDs of current attractors to the ones
```
awage, line 35:
```suggestion
Returns the replacement map, mapping IDs of current attractors to the ones
```

src/matching/matching_interface.jl
```diff
@@ -0,0 +1,193 @@
...
+they match to in previous attractors, according to `matcher`, and also given
+then dynamical system, current and previous parameter values.
```
awage, line 37:
```suggestion
the dynamical system, current and previous parameter values.
```

src/matching/matching_interface.jl
Datseris: Julia style guide is that docstrings are written in imperative form: "do this, give that".

src/matching/sssdistance.jl
```diff
@@ -0,0 +1,123 @@
...
+
+Sets (in particular, their corresponding IDs) are then matched according to this distance.
+First, all possible ID pairs (old, new) are sorted according to the distance of their corresponding sets.
+The pair with smallest distance is matched. IDs in matched pairs are removed from the
+matching pool to ensure a unique mapping. Then, the next pair with least
+remaining distance is matched, and the process repeats until all pairs are exhausted.
+
+Additionally, you can provide a `threshold` value. If the distance between two sets
+is larger than this `threshold`, then it is guaranteed that the sets will get assigned
+different ID in the replacement map (the next available integer).
```
awage, line 43: There something I am not understanding well. Suppose the set  `a+[i]`  has a minimal distance to all sets in `a-` above the threshold. Only the set `a+` gets a new Id, right? From the docstring it seems that both sets `a+` and `a-` get a different id.

Datseris: In the docstring of `replacement_map` it is clarified that the only dictionary that is adjusted is `a+`. `a-` is never adjusted. Above "all sets" refers to all sets in `a+`.

Datseris: actually okay I made it much more clear:

Additionally, you can provide a `threshold` value. If the distance between two sets
is larger than this `threshold`, then it is guaranteed that the two sets will get assigned
different ID in the replacement map, and hence, the set of `a₊` gets the next available integer.

src/continuation/continuation_ascm_generic.jl
```diff
@@ -0,0 +1,235 @@
...
+
+### Step 1 - Seeding initial conditions
+
+Initial conditions can be seeded from previously found attractors.
+This is controlled by the `seeding` keyword, which must be a function that given
+a `StateSpaceSet` (an attractor), it returns an iterator of initial conditions.
+By default the first point of an attractor is provided as the only seed.
+
+Seeding can be turned off by providing the dummy function `seeding = A -> []`,
+i.e., it always returns an empty iterator and hence no seeds.
```
awage, line 38:
```suggestion
Seeding can be turned off by providing the dummy function `seeding = A -> []`,
i.e., it always returns an empty iterator and hence no seeds and we skip Step 2.
```

src/continuation/continuation_ascm_generic.jl
```diff
@@ -0,0 +1,235 @@
...
+I.e., exactly as in [`basins_fractions`](@ref).
+Naturally, during this step new attractors may be found, besides those found
+using the "seeding from previous attractors".
+
+### Step 4 - Matching
+
+Normally the ID an attractor gets assigned is somewhat a random integer.
+Therefore, to ensure a logical output of the global continuation process,
+attractors need to be "matched".
+This means: attractor and fractions must have their _IDs are changed_,
```
awage, line 66:
```suggestion
This means: attractor and fractions must have their _IDs changed_,
```

awage:
This is looking very good. The interface is great.

Datseris:
Thanks @awage . @KalelR I've finished the Basin enclosure matcher, which now is indeed a matcher. I departed significantly from your code in #133 , however, I think the overall logic stays similar. The advantage of my approach is that the source code is significantly smaller as it re-uses existing structures. I decided to use `AttractorsViaProximity` to find the "flowing" and "coflowing" attractors. I thought re-using the proximity mapper would give more robust results than deciding to arbitrarily evolve for some `T` and `Ttr` time.

Here is the full source code:

```julia
"""
    MatchByBasinEnclosure(; kw...) <: IDMatcher

A matcher that matches attractors by whether they are enclosed in
the basin of a new attractor or not.

## Keyword arguments

- `ε = nothing`: distance threshold given to [`AttractorsViaProximity`](@ref).
  If `nothing`, it is estimated as half the minimum distance of centroids
  (in contrast to the default more accurate estimation in [`AttractorsViaProximity`](@ref)).
- `Δt = 1, consecutive_lost_steps = 1000`: also given to [`AttractorsViaProximity`](@ref).
  Note that attractors that did not convergen anywhere within this number of steps
  do not get assigned ID -1 as in [`AttractorsViaProximity`](@ref). Rather, they
  get assigned the next available free ID.
- `distance = Centroid()`: metric to estimate distances between state space sets
  in case there are co-flowing attractors, see below.
- `seeding = A -> A[end]`: how to select a point from the attractor to see if
  it is enclosed in the basin of a new attractor.

## Description

An attractor `A₋` is a set in a state space that occupies a particular region
(or, a single point, if it is a fixed point).
This region is always within the basin of said attractor.
When the parameter of the dynamical system is incremented,
the attractors `A₊` in the new parameter have basins that may have changed in shape and size.

The previous attractor `A₋` is "matched" (i.e., has its ID changed)
to a new attractor `A₊` attractor if `A₋` is located inside the basin of attraction of `A₊`.
To see if `A₋` is in the basin of `A₊`, we first pick a point from `A₊` using the `seeding`
keyword argument. By default this is the last point on the attractor, but it could be anything
else, including the centroid of the attractor (`mean(A)`).
This point is given as an initial condition to an [`AttractorsViaProximity`](@ref) mapper
that maps initial conditions to the `₊` attractors with threshold `ε`
(by default estimated automatically, see [`AttractorsViaProximity`](@ref)).

There can be the situation where multiple `₋` attractors get matched to the same `₊`
attractor, which we call "coflowing attractors". In this scenario matching is prioritized
for the `₋` attractor that is closest to the `₊` in terms of state space set distance,
which is estimated with the `distance` keyword, which can be anything
[`setsofsets_distances`](@ref) accepts. The closes `₋` coflowing attractor
gets assigned the same ID as the `₊` one, while the rest get different unique IDs.

Basin enclosure is a concept similar to "basin instability" in [Ritchie2023](@cite).
"""
@kwdef struct BasinEnclosure{E, D} <: IDMatcher
    ε::E = nothing
    distance::D = Centroid()
    seeding::S = A -> A[end]
    Δt::Float64 = 1
    consecutive_lost_steps::Int = 1000
end

function _match_attractors(
        current_attractors, prev_attractors, matcher::MatchByBasinEnclosure,
        mapper, p, pprev
    )
    ds = referenced_dynamical_system(mapper)
    if matcher.ε === nothing
        e = ε_from_centroids(attractors)
    else
        e = matcher.ε
    end
    proximity = AttractorsViaProximity(ds, current_attractors, e; horizon_limit = Inf)
    rmap = Dict(k => proximity(matcher.seeding(A)) for (k, A) in prev_attractors)
    # we now process the replacement map `rmap` for co-flowing or diverged attractors.
    next_id = next_free_id(current_attractors, prev_attractors)
    # take care of diverged attractors
    for (old_ID, new_ID) in rmap
        if new_ID < 0 # diverged attractors get -1 ID.
            rmap[old_ID] = next_id
            next_id += 1
        end
    end
    # next, take care of co-flowing attractors
    if unique(values(rmap)) != length(rmap) # a value is repeated
        # Do coflowing and assign `next_id` to the least distant coflowing
    end
    return rmap
end

function ε_from_centroids(attractors)
    distances = setsofsets_distances(attractors, attractors, Centroid())
    alldists = sort!(vcat([collect(values(d)) for (k,d) in distances]...))
    filter!(!iszero, alldists)
    return minimum(alldists)/2
end
```

The only thing left is to implement the co-flowing logic.

Datseris:
One thing you didn't explain in #133 is how you handle the "other" attractors in the coflowing case. The closest attractor gets the ID of the new one it flows to. What about the others though? What ID do they get? Do they get the next free available integer? Or do we check if there are some new attractors where no previous attractor has flowed to, and assign this ID by proximity? I think it is easier to assign the next available integer, but we likely will have to test this with your Kuramoto models.

For now, I will leave this as it is, and let you play around with this implementation by adjusting what happens inside the "co flowing attractors" `if` clause. It would be fantastic if we could have this done before JuliaCon 2024.

Now, I will focus on ensuring the tests pass and merge in this PR and tag a new release.

KalelR:
> For now, I will leave this as it is, and let you play around with this implementation by adjusting what happens inside the "co flowing attractors" `if` clause. It would be fantastic if we could have this done before JuliaCon 2024.

Sorry I took so long.
So let's say that at parameter p1 there two attractors, A1- and A2-. At parameter p+>p- there is only one attractor, Ak+. Trajectories starting at A1- and A2- (with p = p+) flow onto Ak+. They are thus coflowing attractors. But the distance d(A1-, Ak+) < d(A2-, Ak+). Therefore, we match Ak+ with A1-, thus labelling Ak+ as A1+. In the continuation, we would see A1- continue onto A1+ and A2- cease to exist. This is the behavior I originally intended, and the code is already doing this (with some modifications I'm introducing in a PR).

PR #134, opened 2024-07-02
Finish tutorial and matcher interface

Datseris:
This PR finishes the tutorial and also truly finishes the matcher interface. The interface is based on the `matching_map` function, which is a rename of the `replacement_map` function. See the docstring of `IDMatcher` and `matching_map` for the new interface.

The PR adds a comparison also with BifurcationKit.jl.

In this PR I've also implemented the desired `MatchByBasinEnclosure`. It is part of the interface now so it is exported and added to the docs. I've also tested that it "runs" with the tutorial example. But, I have no good system at the moment to test it with. If I use the system of the tutorial, and after I match the normal way with the tutorial I try again by doing:

```julia
matcher = MatchByBasinEnclosure()
attractors_cont2 = deepcopy(attractors_cont)
match_sequentially!(attractors_cont2, matcher; ds, prange, pidx)
fig = plot_attractors_curves(
	attractors_cont2, A -> minimum(A[:, 1]), prange,
)
```
then I get 

image

notice how when we go into the first bistable regime there is something wonky (a black triangle is high in y axis), and when we go to the second bistable regime the attractor labels alternate all the time. I don't know if this is correct for this system because the basins are super duper fractal and thus everything all the time fluctuates a lot, or there is something wrong in my logic. In any case, I'd appreciate some help testing this new matcher.


Datseris:
I am presenting this tomorrow so I am merging it in. It would be fantastic to get tests for the new basin enclosure matcher @KalelR , but this can be done when you have some time. Would also be fantastic to have an Example with your kuramoto system that shows that using other matchers doesn't work as well.

Issue #137, opened 2024-07-22
mapper(u0)  not defined for AttractorsViaFeaturizing


awage:
In the docstring of `AttractorsViaFeaturizing` we have: 
``` julia
This `mapper` also allows the syntax `mapper(u0)` but only if the `grouping_config`
is _not_ `GroupViaClustering`.
```
but the function is not implemented: 
```julia
   function dumb_map(z, p, n)
        x, y = z
        r = p[1]
        if r < 0.5
            return SVector(0.0, 0.0)
        else
            if x ≥ 0
                return SVector(r, r)
            else
                return SVector(-r, -r)
            end
        end
    end

    r = 1.0
    ds = DeterministicIteratedMap(dumb_map, [0., 0.], [r])
    u0s = [1 => [r, r], 2 => [-r, -r]] # template ics

    xg = yg = range(-2.0, 2.0; length=100)
    grid = (xg, yg)
    expected_fs_raw = Dict(1 => 0.5, 1 => 0.5)
    featurizer(A, t) = SVector(A[1][1])

function features_from_u(u)
    A, t = trajectory(ds, 100, u; Ttr = 1000, Δt = 1)
featurizer(A, t)
end
templates = Dict(k => features_from_u(u) for (k, u) in u0s)
config = GroupViaNearestFeature(templates)
mapper = AttractorsViaFeaturizing(ds, featurizer, config; Ttr=500)
mapper([1., 0.])
```
returns: 

ERROR: MethodError: objects of type AttractorsViaFeaturizing{DeterministicIteratedMap{false, SVector{2, Float64}, 2, 
typeof(dumb_map), Vector{Float64}}, GroupViaNearestFeature{1, Float64, NearestNeighbors.KDTree{SVector{1, Float64}, 
Euclidean, Float64, SVector{1, Float64}}, Vector{Int64}}, typeof(featurizer), Int64, 2, Float64} **are not callable**

I will propose a PR to solve this. It is not crucial but it a convenience function that is helpful.

Datseris:
Sorry @awage the docstring is wrong. With `GroupViaNearestFeature` it is not possible to do `mapper(u0)`. There is simply no other nearest feature to map to, so it doesn't make sense. We just need to correct the docstring.

awage:
It is possible since you give the templates to `GroupViaNearestFeature.` : 

[...]

PR #138, opened 2024-07-22
Add the mapper function for AttractorsViaFeaturinzing

awage:
Problem described in #137. I added a simple function. It works fine with `GroupViaNearestFeature`, however I don't know what is the behavior for other `GroupConfig.` I suppose it will throw an error since `feature_to_group` is not defined for some GroupConfig.

### DIFF
--- a/src/mapping/grouping/attractor_mapping_featurizing.jl
+++ b/src/mapping/grouping/attractor_mapping_featurizing.jl
@@ -166,6 +166,12 @@ function extract_features_single(mapper, ics; show_progress = true, N = 1000)
     return feature_vector
 end
 
+# Mapper function that works with GroupViaNearestFeature
+function (mapper::AttractorsViaFeaturizing)(u0)
+   f = extract_features_single(mapper,[u0]) 
+   return feature_to_group(f[1],mapper.group_config) 
+end
+
 # TODO: We need an alternative to deep copying integrators that efficiently
 # initializes integrators for any given kind of system. But that can be done
 # later in the DynamicalSystems.jl 3.0 rework.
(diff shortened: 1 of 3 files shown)


awage:
A test is needed, I will add one soon.

src/mapping/grouping/attractor_mapping_featurizing.jl
```diff
@@ -166,6 +166,12 @@ function extract_features_single(mapper, ics; show_progress = true, N = 1000)
     return feature_vector
 end
 
+# Mapper function that works with GroupViaNearestFeature
+function (mapper::AttractorsViaFeaturizing)(u0)
+   f = extract_features_single(mapper,[u0]) 
+   return feature_to_group(f[1],mapper.group_config)
```
Datseris, line 172:
```suggestion
   f = extract_features_single(mapper,[u0]) 
   return feature_to_group(f[1],mapper.group_config) 
```

Here careful with styling, add space between arguments.

src/mapping/grouping/attractor_mapping_featurizing.jl
```diff
@@ -166,6 +166,12 @@ function extract_features_single(mapper, ics; show_progress = true, N = 1000)
     return feature_vector
 end
 
+# Mapper function that works with GroupViaNearestFeature
```
Datseris, line 169: This comment is missleading. This function also works with the histogram grouping. That's why I don't like comments above functions. We should remove this comment.

Datseris (commented):
Yeah this is fine. Calling this with some groupings will error, but that is fine.

Datseris (approved):
All good, also increment patch version in Project.toml so that we can release a new version immediatelly.

Datseris:
@awage did you tag a new release after you merged this in? WHoever merges must also tag the release. I Tag releases by going on https://juliahub.com/ui/Packages and clicking "register package" in the top right.

awage:
No I didn’t. I will do it in a couple of hours. If you need it right now,
please tag the release.

PR #141, opened 2024-08-02
Improve docstring and performance of `basin_entropy`

Datseris:
I was reading it and I couldn't figure out how the probabilities were computed so I added the equation from the paper. I also looked at the source code that could be optimized. In the end the optimizATIONS don't bring much performance 
```
julia> @btime basin_entropy($basins)
  1.700 ms (5200 allocations: 1.68 MiB)
(0.1944289341841219, 0.7070143061240797)

julia> @btime basin_entropy2($basins)
  1.576 ms (2400 allocations: 212.50 KiB)
(0.1944289341841219, 0.7070143061240797)
```
but at least they reduce the allocated memory 16x fold.

### DIFF
--- a/src/basins/fractality_of_basins.jl
+++ b/src/basins/fractality_of_basins.jl
@@ -42,7 +42,7 @@ function basin_entropy(basins::AbstractArray{Int, D}, ε::Int = 20) where {D}
     return basin_entropy(basins, es)
 end
 
-function basin_entropy(basins::AbstractArray{Int, D}, es::Dims{D})
+function basin_entropy(basins::AbstractArray{Int, D}, es::Dims{D}) where {D}
     Sb = 0.0; Nb = 0
     εranges = map((d, ε) -> 1:ε:d, size(basins), es)
     box_iterator = Iterators.product(εranges...)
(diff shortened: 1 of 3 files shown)


src/basins/fractality_of_basins.jl
```diff
@@ -3,20 +3,30 @@ export uncertainty_exponent, basins_fractal_dimension, basins_fractal_test, basi
...
+as [Daza2016](@cite)
+
+```math
+S_b = \\tfrac{1}{N}\\sum_{i=1}^{N}\\sum_{j=1}^{m_i}-p_{ij}\\log(p_{ij})
+```
+where ``m`` is the number of unique IDs (integers of `basins`) in box ``i``
+and ``p_{ij}`` is the relative frequency (probability) to obtain ID ``j``
+in the ``i`` box (simply the count of IDs ``j`` divided by the total in the box).
+
+`Sbb` is the boundary basin entropy `Sbb`.
```
awage, line 23:
```suggestion
`Sbb` is the boundary basin entropy.
```

awage (approved):

awage:
Better docs, better performances. All good!!

Datseris:
@awage I think our code is wrong. or at least, it is not aligning with the documentation.

We say that `Sbb` is the same as `Sb` but it is averaged over only boxes that are on boundaries.

But that is not true. `Sbb` is `Sb` divided by the number of boxes on boundaries. But it is not estimated _only_ on boxes on boundaries. It is still estimated on all boxes and it is only normalized differently. Is this intended?

Datseris:
right of course forget what i said entropy is 0 anyways in the other cases.

Issue #143, opened 2024-08-05
Problem with matching by set distance


KalelR:
I'd like some help in figuring out what's going wrong in a synthetic example I created. I'm trying to apply global_continuation using AttractorsViaFeaturizing, first with a matching by set distance. The system's real behavior is as described in the left column of the figure below.

![matching_by_flow_example-method-2-dummy_multistable_equilibrium!](https://github.com/user-attachments/assets/09895543-9fcf-4d29-b74a-82a1d6532189)

The first attractor, which starts at position 0, moves to -5 for params 3 and 3.5, but moves back to 0 at param>=4.  Matching by centroid misses this behavior, and 'incorrectly' matches the attractors, as shown in the right column.

I had run these results before the reformulation of `global_continuation` and everything. Now I am getting an error:
![image](https://github.com/user-attachments/assets/58c286f5-0b1b-47ad-864a-e69d9249490c)

There are missing values of the basin fractions. As far as I understand, the problem is in the `rmap` that you get. This simple example shows what's happening:
```julia 
fs = Dict(4 => 0.4097706879361914, 2 => 0.19042871385842472, 3 => 0.18843469591226322, 1 => 0.21136590229312063)
rmap = Dict(2 => 3) #error
# rmap = Dict(2 => 3, 3=>2) #fine
 
swap_dict_keys!(fs, rmap)
```

The rmap for those parameters is missing one entry (e.g. 3=>2). This makes the swap function drop one of the keys. I am not sure why this is happening, but I know that removing line 112 of `sssdistance.jl`, which is 
```julia 
    filter!(p -> p.first ≠ p.second, rmap)
```

solves it, because then rmap has 2=>3 and 3=>2. I don't know why, though. 
Could you please confirm this is not just on my machine? I've updated everything now.

If you don't want to run the continuation, it all boils down to these guys: 
```julia
pcurve = [[1 => 1.0], [1 => 1.5], [1 => 2.0], [1 => 2.5], [1 => 3.0], [1 => 3.5], [1 => 4.0], [1 => 4.5], [1 => 5.0]]
[...]

PR #144, opened 2024-08-06
Fix matching by SSSet distance

KalelR:
As discussed in #143, filtering out repetitive (e.g., 4=>4) entries in rmap can lead to problems when key_retraction is used. A simple fix that works is to remove the filter line, as I'm doing here.

I've also added the test I was doing when I found this error. Without the filter, the code works great, resulting in the right column in this figure 

matching_by_flow_example-method-2-dummy_multistable_equilibrium!

Note that this is not the "actual" behavior of the system, as by construction the attractor at 0 moves to -5, as shown on the left column. This is however a consequence of matching by distance in state space, not a problem with the code.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "Attractors"
 uuid = "f3fd9213-ca85-4dba-9dfd-7fc91308fec7"
 authors = ["George Datseris <datseris.george@gmail.com>", "Kalel Rossi", "Alexandre Wagemakers"]
 repo = "https://github.com/JuliaDynamics/Attractors.jl.git"
-version = "1.19.1"
+version = "1.19.2"
 
 [deps]
 BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
--- a/src/matching/sssdistance.jl
+++ b/src/matching/sssdistance.jl
@@ -108,8 +108,6 @@ function _matching_map_distances(keys₊, keys₋, distances::Dict, threshold;
         end
     end
 
-    # the final step is to filter out equivalent mappings where key and value are the same
-    filter!(p -> p.first ≠ p.second, rmap)
     return rmap
 end
 
--- a/test/continuation/matching_attractors.jl
+++ b/test/continuation/matching_attractors.jl
@@ -39,53 +39,20 @@ end
     end
 end
 
-@testset "synthetic_multistable_continuation" begin 
-    function dummy_multistable_equilibrium!(dx, x, p, n)
-        r = p[1]
-        num_max_atts = 5
-        x_max_right = 10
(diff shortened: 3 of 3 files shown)


Datseris (changes requested):
The current test added is unecessarily complex. This is not problem with continuation, it is a problem with matching, only,. There is no need to define a dynamical system or make a continuation.

Why don't you just use this code?

```julia
using Attractors

attractors_cont_simple = Dict{Int64, SVector{1, Float64}}[Dict(1 => [0.0]), Dict(1 => [0.0]), Dict(2 => [2.0], 1 => [0.0]), Dict(2 => [0.0], 1 => [2.0]), Dict(2 => [2.0], 3 => [4.0], 1 => [-5.0]), Dict(2 => [4.0], 3 => [-5.0], 1 => [2.0]), Dict(4 => [6.0], 2 => [0.0], 3 => [2.0], 1 => [4.0]), Dict(4 => [4.0], 2 => [0.0], 3 => [2.0], 1 => [6.0]), Dict(5 => [8.0], 4 => [6.0], 2 => [0.0], 3 => [2.0], 1 => [4.0])]
attractors_cont = [Dict(k=>StateSpaceSet(Vector(v)) for (k,v) in atts) for atts in attractors_cont_simple]


mapped_atts = deepcopy(attractors_cont)
matcher = MatchBySSSetDistance()
rmaps = match_sequentially!(mapped_atts, matcher)

fractions_cont = [Dict(1 => 1.0), Dict(1 => 1.0), Dict(2 => 0.8091908091908092, 1 => 0.19080919080919082), Dict(2 => 0.1906187624750499, 1 => 0.8093812375249501), Dict(2 => 0.18862275449101795, 3 => 0.6207584830339321, 1 => 0.1906187624750499), Dict(2 => 0.6211365902293121, 3 => 0.19042871385842472, 1 => 0.18843469591226322), Dict(4 => 0.4097706879361914, 2 => 0.19042871385842472, 3 => 0.18843469591226322, 1 => 0.21136590229312063), Dict(4 => 0.21115537848605578, 2 => 0.1902390438247012, 3 => 0.18824701195219123, 1 => 0.4103585657370518), Dict(5 => 0.21613545816733068, 4 => 0.1942231075697211, 2 => 0.1902390438247012, 3 => 0.18824701195219123, 1 => 0.21115537848605578)]

mapped_fracs = deepcopy(fractions_cont)
match_sequentially!(mapped_fracs, rmaps)

```

Furthermore, why are the basins such complex numbers instead of just `0.1, 0.2, ...` etc.? You can make them programmatically from the `attractors` by equi-distributing the number 1 to all attractors.

Datseris (approved):
don't forget to increment patch version in Project.toml

PR #146, opened 2024-08-06
Fixing BasinEnclosure matcher

KalelR:
I have updated the code for the BasinEnclosure method with two main fixes:
1. get the logic right for `_grouped_flows_`
2. fix missing entries in rmaps. When dealing with coflowing attractors we modify `rmap`. But this might leading to missing keys. In this case, the simplest solution I found was to add those missing keys back with the next available integer key in `rmap`. 

I've tested this with a synthetic multistable system, whose code I've also added to the continuation test suite. The correct behavior is as shown on the left column. Note that the attractor at 0.0 moves to -5.0, with the initial conditions following it. The algorithm identifies and correctly matches the attractors. Note that the other matcher, which relies on distances in state space, misses this. Because 0 is closer to 4 than to -5, it matches 0 to 4, and considers -5 as a new attractor.

georgescode_matching_by_flow_example-method-2-dummy_multistable_equilibrium!

I've made some more tests using a similar system, in which the attractors are oscillating. The results are the same.
georgescode_matching_by_flow_example-method-2-dummy_multistable_sinusoidal!

I've also played with some different synthetic scenarios, and the code also worked there. Further, I tested on the example we used for continuation with featurizing in the magnetic pendulum, and the code also worked there.

I will test this on the Kuramoto networks, but not now. It is somewhat involved for the important testcases (with N=100) and I have to work on other projects first.

### DIFF
--- a/src/matching/basin_enclosure.jl
+++ b/src/matching/basin_enclosure.jl
@@ -54,25 +54,6 @@ Basin enclosure is a concept similar to "basin instability" in [Ritchie2023](@ci
     consecutive_lost_steps::Int = 1000
 end
 
-function next_key(dict)
-    allkeys = collect(values(dict))
-    maxkey = maximum(allkeys)
-    return maxkey+1
-end
-
-# missing key: a key that is in a₊ but not in keys(rmap);
-# when keys are swapped, this missing key might be lost
-# to fix it, I add the missing key to rmap, pointing to the next available key
-function add_missing_keys_rmap!(rmap, current_attractors)
-    allkeys = collect(keys(current_attractors))
-    usedkeys = collect(keys(rmap))
-    missingkeys = [k for k in allkeys if !(k ∈ usedkeys)]
-    for missingkey in sort(missingkeys)
-        rmap[missingkey] = next_key(rmap)
-    end
-    return
-end
-
 function matching_map(
         current_attractors, prev_attractors, matcher::MatchByBasinEnclosure;
         ds, p, pprev = nothing, next_id = next_free_id(current_attractors, prev_attractors)
@@ -104,11 +85,13 @@ function matching_map(
         end
     end
     # next up are the co-flowing attractors
-    grouped_flows = _grouped_flows(flow)
+    allnewids = keys(current_attractors) #needed because flow only includes new ids that prev attractors flowed onto
+    grouped_flows = _grouped_flows(flow, allnewids) #map all current ids to prev ids that flowed onto them (if the att is new, no prev ids flowed, and the corresponding entry is empty)
+
     # notice the keys of `grouped_flows` are new IDs, same as with `rmap`.
     for (new_ID, old_flowed_to_same) in grouped_flows
-        if length(old_flowed_to_same) == 0
-            continue # none of the old IDs converged to the current `new_ID`
+        if length(old_flowed_to_same) == 0 # none of the old IDs converged to the current `new_ID`
+            rmap[new_ID] = next_id #necessary to make rmap complete and avoid skipping keys
         elseif length(old_flowed_to_same) == 1
             rmap[new_ID] = only(old_flowed_to_same)
         else # need to resolve coflowing using distances
@@ -123,8 +106,6 @@ function matching_map(
             rmap[new_ID] = old_ID
         end
     end
-    add_missing_keys_rmap!(rmap, current_attractors)
-
     return rmap
 end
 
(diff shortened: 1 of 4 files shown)


src/matching/basin_enclosure.jl
```diff
@@ -54,6 +54,25 @@ Basin enclosure is a concept similar to "basin instability" in [Ritchie2023](@ci
     consecutive_lost_steps::Int = 1000
 end
 
+function next_key(dict)
+    allkeys = collect(values(dict))
+    maxkey = maximum(allkeys)
+    return maxkey+1
+end
```
Datseris, line 61: You have to use the existing infrastructure with `next_id`. It is already passed to the `matching_map` function. Please see how `next_id` is handled in the `matching_map` implementation for `MatchBySSSetDistance`.

src/matching/basin_enclosure.jl
```diff
@@ -54,6 +54,25 @@ Basin enclosure is a concept similar to "basin instability" in [Ritchie2023](@ci
...
+
+# missing key: a key that is in a₊ but not in keys(rmap);
+# when keys are swapped, this missing key might be lost
+# to fix it, I add the missing key to rmap, pointing to the next available key
+function add_missing_keys_rmap!(rmap, current_attractors)
+    allkeys = collect(keys(current_attractors))
+    usedkeys = collect(keys(rmap))
+    missingkeys = [k for k in allkeys if !(k ∈ usedkeys)]
+    for missingkey in sort(missingkeys)
+        rmap[missingkey] = next_key(rmap)
```
Datseris, line 71: here use the `next_id` mentioned above.

Datseris: The `next_id` input of `matching_map` is given as input to this function.

Datseris (changes requested):
Don't forget to increment patch version in Project.toml.

KalelR:
I've implemented a cleaner, more straightforward logic now to handle these "missing keys" in rmap. Now the loop already includes them and they are not missing. I've also used next_id as you suggested.

src/matching/basin_enclosure.jl
```diff
@@ -104,11 +85,13 @@ function matching_map(
...
-    grouped_flows = _grouped_flows(flow)
+    allnewids = keys(current_attractors) #needed because flow only includes new ids that prev attractors flowed onto
+    grouped_flows = _grouped_flows(flow, allnewids) #map all current ids to prev ids that flowed onto them (if the att is new, no prev ids flowed, and the corresponding entry is empty)
+
     # notice the keys of `grouped_flows` are new IDs, same as with `rmap`.
     for (new_ID, old_flowed_to_same) in grouped_flows
-        if length(old_flowed_to_same) == 0
-            continue # none of the old IDs converged to the current `new_ID`
+        if length(old_flowed_to_same) == 0 # none of the old IDs converged to the current `new_ID`
+            rmap[new_ID] = next_id #necessary to make rmap complete and avoid skipping keys
```
Datseris, line 94: This is not correct because `next_id` is not incremented. Imagine that in `grouped_flows` there is more than one element `old_flowed_to_same` that has length > 0. Then, all `new_ID` will get the same `next_id`. What you have to do is add a new line after `rmap[new_ID] = next_id` that simply does `next_id += 1`.

src/matching/basin_enclosure.jl
```diff
@@ -149,7 +130,8 @@ function ε_from_centroids(attractors::AbstractDict)
 end
 
 # group flows so that all old IDs that go to same new ID are in one vector
-function _grouped_flows(flows) # separated into
-    grouped = Dict(v2=>[k for (k,v) in flows if v==v2] for (k2, v2) in flows)
+# i.e., new id => [prev ids that flowed to new id]
+function _grouped_flows(flows, allnewids) # separated into
+    grouped = Dict(newid=>[k for (k,v) in flows if v==newid] for newid in allnewids)
```
Datseris, line 135: thanks for adding this source code clarity!

Datseris (changes requested):

src/matching/basin_enclosure.jl
KalelR: true!

KalelR: true!

Datseris (approved):

Datseris:
@KalelR PR can't be merged, there are conflicts, please resolve them as well!

PR #148, opened 2024-09-01
Stagger and step method for approximating chaotic saddles

awage:
This is an implementation the stagger-and-step method [^Sweet2001] and [^Sala2016] to approximate the invariant 
non-attracting set governing the chaotic transient dynamics of a system, namely the stable manifold of a chaotic saddle. 

The method relies on the stagger-and-step algorithm that 
computes points close to the saddle that escapes in a time 
`T(x_n) > Tm`. The function `T` represents the escape time 
from a region defined by the user (see the argument 
`isinside`).

Given the dynamical mapping `F`, if the iteration `x_{n+1} = F(x_n)` 
respects the condition `T(x_{n+1}) > Tm` we accept 
this next point, this is the _step_ part of the method. If 
not, the method search randomly the next point in a 
neighborhood following a given probability distribution, this 
is the _stagger_ part. This part sometimes fails to find a new 
candidate and a new starting point of the trajectory is chosen 
within the defined region. See the keyword argument 
`stagger_mode` for the different available methods.   

The method produces a pseudo-trajectory of `N` points δ-close
to the stable manifold of the chaotic saddle. 

The original method is implemented as well as an improvement by Sala et. al. 

I added an example file for the docs.

### DIFF
--- a/src/boundaries/example_stagger.jl
+++ b/src/boundaries/example_stagger.jl
@@ -1,0 +1,30 @@
+using Attractors
+using CairoMakie
+
+# Coupled Hénon maps
+function F!(du, u ,p, n)
+    x,y,u,v = u
+    A = 3; B = 0.3; C = 5.; D = 0.3; k = 0.4; 
+    du[1] = A - x^2 + B*y + k*(x-u)
+    du[2] = x
+    du[3] = C - u^2 + D*v + k*(u-x)
+    du[4] = u
+    return 
+end
+
+# The region should not contain any attractors. 
+R_min = [-4; -4.; -4.; -4.] 
+R_max = [4.; 4.; 4.; 4.]
+
+
+# Initial box and initial condition.
+sampler, isinside = statespace_sampler(HRectangle(R_min,R_max))
+x0 = sampler()
+df = DeterministicIteratedMap(F!, x0) 
+xi = stagger_trajectory!(df, x0, 30, isinside; δ₀= 2.) 
+@show Tp = Attractors.escape_time!(df, xi, isinside)
+
+v = stagger_and_step!(df, x0, 10000, isinside; stagger_mode = :adaptive, δ = 1e-4, Tm = 10, max_steps = Int(1e5), δ₀ = 2.) 
+v = hcat(v...)'
+scatter(v[:,1], v[:,3]; markersize = 3)
+
--- a/src/boundaries/stagger_and_step.jl
+++ b/src/boundaries/stagger_and_step.jl
@@ -5,19 +5,20 @@ using ProgressMeter
 
 
 """
-    stagger_trajectory!(ds, x0, isinside; kwargs...) -> xi
+    stagger_trajectory!(ds, x0, Tm, isinside; kwargs...) -> xi
 
 This function returns a point `xi` which _guarantees_ `T(xi) > 
 Tm` with a random walk search around the initial coordinates 
-`x0`. There is no guarantee to find such point, the parameters 
+`x0`. `T(xi)` is the escape time of the initial condition `xi`.  
+There is no guarantee to find such point, the parameters 
 must be adjusted to find the suitable point. This is an .
-auxiliary function for [`stagger_and_step`](@ref). Keyword arguments and definitions are identical for both 
-functions. 
+auxiliary function for [`stagger_and_step`](@ref). Keyword arguments and 
+definitions are identical for both functions. 
 
-The initial search radius is much bigger, `δ₀ = 1.` by default.
+The initial search radius `δ₀` is big, `δ₀ = 1.` by default.
 
 """
-function stagger_trajectory!(ds, x0, isinside; δ₀ = 1., Tm = 30, stagger_mode = :exp, max_steps = Int(1e5), f = 1.1)
+function stagger_trajectory!(ds, x0, Tm, isinside; δ₀ = 1., stagger_mode = :exp, max_steps = Int(1e5), f = 1.1)
     T = escape_time!(ds, x0, isinside)
     xi = deepcopy(x0) 
     while T < Tm  # we must have T ≥ Tm at each step 
@@ -117,7 +118,7 @@ enough to find a suitable initial.
 function stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; δ = 1e-10, Tm  = 30, 
     f = 1.1, max_steps = Int(1e5), stagger_mode = :exp, δ₀ = 1.)
 
-    xi = stagger_trajectory!(ds, x0, isinside; δ₀, Tm, stagger_mode = :unif, max_steps) 
+    xi = stagger_trajectory!(ds, x0, Tm, isinside; δ₀, stagger_mode = :unif, max_steps) 
     v = Vector{Vector{Float64}}(undef,N)
     v[1] = xi
 @showprogress   for n in 1:N
@@ -127,7 +128,7 @@ function stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; 
             xp, Tp = get_stagger!(ds, xi, δ, Tm, isinside; stagger_mode, max_steps, f)
             # The stagger step may fail. We reinitiate the algorithm from a new initial condition.
             if Tp < 0
-                xp = stagger_trajectory!(ds, x0, isinside; δ₀, Tm, stagger_mode = :exp, max_steps, f) 
+                xp = stagger_trajectory!(ds, x0, Tm, isinside; δ₀, stagger_mode = :exp, max_steps, f) 
                 δ = 0.1
             end
             set_state!(ds,xp)
--- a/src/boundaries/stagger_and_step.jl
+++ b/src/boundaries/stagger_and_step.jl
@@ -5,7 +5,7 @@ using ProgressMeter
 
 
 """
-    stagger_trajectory!(x0 ,ds, isinside; kwargs...) -> xi
+    stagger_trajectory!(ds, x0, isinside; kwargs...) -> xi
 
 This function returns a point `xi` which _guarantees_ `T(xi) > 
 Tm`. This is an auxiliary function for [`stagger_and_step`](@
@@ -15,11 +15,11 @@ functions.
 The initial search radius is much bigger, `δ = 1.` by default.
 
 """
-function stagger_trajectory!(x0 ,ds, isinside; δ = 1., Tm = 30, stagger_mode = :exp, max_steps = Int(1e5), f = 1.1)
+function stagger_trajectory!(ds, x0, isinside; δ = 1., Tm = 30, stagger_mode = :exp, max_steps = Int(1e5), f = 1.1)
     T = escape_time!(x0, ds, isinside)
     xi = deepcopy(x0) 
     while T < Tm  # we must have T ≥ Tm at each step 
-        xi, T = get_stagger!(xi, ds, δ, T, isinside; f, stagger_mode, max_steps)
+        xi, T = get_stagger!(ds, xi, δ, T, isinside; f, stagger_mode, max_steps)
         if T < 0
             error("Cannot find a stagger trajectory. Choose a different starting point or search radius δ.")
         end 
@@ -30,7 +30,7 @@ end
 
 
 """
-    stagger_and_step!(x0, ds::DynamicalSystem , N::Int, isinside::function; kwargs...) -> trajectory
+    stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; kwargs...) -> trajectory
 
 This function implements the stagger-and-step method 
 [^Sweet2001] and [^Sala2016] to approximate the invariant 
@@ -108,20 +108,20 @@ to the stable manifold of the chaotic saddle.
 [^Sweet2001]: D. Sweet, *et al.*, Phys. Rev. Lett. **86**, pp 2261  (2001)
 [^Sala2016]: M. Sala, *et al.*, Chaos **26**, pp 123124 (2016)
 """
-function stagger_and_step!(x0 ,ds, N, isinside; δ = 1e-10, Tm  = 30, 
+function stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; δ = 1e-10, Tm  = 30, 
     f = 1.1, max_steps = Int(1e5), stagger_mode = :exp)
 
-    xi = stagger_trajectory!(x0, ds, isinside; δ = 1., Tm, stagger_mode = :exp, max_steps) 
+    xi = stagger_trajectory!(ds,x0, isinside; δ = 1., Tm, stagger_mode = :exp, max_steps) 
     v = Vector{Vector{Float64}}(undef,N)
     v[1] = xi
 @showprogress   for n in 1:N
         if escape_time!(xi, ds, isinside) > Tm
             set_state!(ds, xi)
         else
-            xp, Tp = get_stagger!(xi, ds, δ, Tm, isinside; stagger_mode, max_steps, f)
+            xp, Tp = get_stagger!(ds, xi, δ, Tm, isinside; stagger_mode, max_steps, f)
             # The stagger step may fail. We reinitiate the algorithm from a new initial condition.
             if Tp < 0
-                xp = stagger_trajectory!(x0, ds, isinside; δ = 1., Tm, stagger_mode = :exp, max_steps, f) 
+                xp = stagger_trajectory!(ds, x0, isinside; δ = 1., Tm, stagger_mode = :exp, max_steps, f) 
                 δ = 0.1
             end
             set_state!(ds,xp)
@@ -172,7 +172,7 @@ end
 # This function searches a new candidate in a neighborhood of x0 
 # with a random search depending on some distribution. 
 # If the search fails it returns a negative time.
-function get_stagger!(x0, ds, δ, Tm, isinside; max_steps = Int(1e6), f = 1.1, stagger_mode = :exp, verbose = false)
+function get_stagger!(ds, x0, δ, Tm, isinside; max_steps = Int(1e6), f = 1.1, stagger_mode = :exp, verbose = false)
 
     Tp = 0; xp = zeros(length(x0)); k = 1; 
     T0 = escape_time!(x0, ds, isinside)
--- a/src/boundaries/stagger_and_step.jl
+++ b/src/boundaries/stagger_and_step.jl
@@ -8,20 +8,22 @@ using ProgressMeter
     stagger_trajectory!(ds, x0, isinside; kwargs...) -> xi
 
 This function returns a point `xi` which _guarantees_ `T(xi) > 
-Tm`. This is an auxiliary function for [`stagger_and_step`](@
-ref). Keyword arguments and definitions are identical for both 
+Tm` with a random walk search around the initial coordinates 
+`x0`. There is no guarantee to find such point, the parameters 
+must be adjusted to find the suitable point. This is an .
+auxiliary function for [`stagger_and_step`](@ref). Keyword arguments and definitions are identical for both 
 functions. 
 
-The initial search radius is much bigger, `δ = 1.` by default.
+The initial search radius is much bigger, `δ₀ = 1.` by default.
 
 """
-function stagger_trajectory!(ds, x0, isinside; δ = 1., Tm = 30, stagger_mode = :exp, max_steps = Int(1e5), f = 1.1)
-    T = escape_time!(x0, ds, isinside)
+function stagger_trajectory!(ds, x0, isinside; δ₀ = 1., Tm = 30, stagger_mode = :exp, max_steps = Int(1e5), f = 1.1)
+    T = escape_time!(ds, x0, isinside)
     xi = deepcopy(x0) 
     while T < Tm  # we must have T ≥ Tm at each step 
-        xi, T = get_stagger!(ds, xi, δ, T, isinside; f, stagger_mode, max_steps)
+        xi, T = get_stagger!(ds, xi, δ₀, T, isinside; f, stagger_mode, max_steps)
         if T < 0
-            error("Cannot find a stagger trajectory. Choose a different starting point or search radius δ.")
+            error("Cannot find a stagger trajectory. Choose a different starting point or search radius δ₀.")
         end 
             
     end
@@ -92,7 +94,7 @@ to the stable manifold of the chaotic saddle.
       distribution in a random direction `u` around the current
       `x` such that `x_c = x + u*r`. `r = 10^-s` with `s` taken
       from a uniform distribution in [-15, δ]. This mode fails
-      often but stills manage to provide long enough stretch of
+      often but still manage to provide long enough stretch of
       trajectories. 
 
     * `:unif`: The next candidate is `x_c = x + u*r` with `r` 
@@ -104,24 +106,28 @@ to the stable manifold of the chaotic saddle.
       when it succeeds.    
 * `f = 1.1`: It is the free parameter for the adaptive stagger
   method. 
-
+* `δ₀ = 1.`: This is the radius for the first stagger 
+trajectory search. The algorithm looks for a point 
+sufficiently close to the saddle before switching to the 
+stagger-and-step routine. The search radius must be large 
+enough to find a suitable initial. 
 [^Sweet2001]: D. Sweet, *et al.*, Phys. Rev. Lett. **86**, pp 2261  (2001)
 [^Sala2016]: M. Sala, *et al.*, Chaos **26**, pp 123124 (2016)
 """
 function stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; δ = 1e-10, Tm  = 30, 
-    f = 1.1, max_steps = Int(1e5), stagger_mode = :exp)
+    f = 1.1, max_steps = Int(1e5), stagger_mode = :exp, δ₀ = 1.)
 
-    xi = stagger_trajectory!(ds,x0, isinside; δ = 1., Tm, stagger_mode = :exp, max_steps) 
+    xi = stagger_trajectory!(ds, x0, isinside; δ₀, Tm, stagger_mode = :unif, max_steps) 
     v = Vector{Vector{Float64}}(undef,N)
     v[1] = xi
 @showprogress   for n in 1:N
-        if escape_time!(xi, ds, isinside) > Tm
+        if escape_time!(ds, xi, isinside) > Tm
             set_state!(ds, xi)
         else
             xp, Tp = get_stagger!(ds, xi, δ, Tm, isinside; stagger_mode, max_steps, f)
             # The stagger step may fail. We reinitiate the algorithm from a new initial condition.
             if Tp < 0
-                xp = stagger_trajectory!(ds, x0, isinside; δ = 1., Tm, stagger_mode = :exp, max_steps, f) 
+                xp = stagger_trajectory!(ds, x0, isinside; δ₀, Tm, stagger_mode = :exp, max_steps, f) 
                 δ = 0.1
             end
             set_state!(ds,xp)
@@ -135,7 +141,7 @@ end
 
 
     
-function escape_time!(x0, ds, isinside) 
+function escape_time!(ds, x0, isinside) 
     x = deepcopy(x0) 
     set_state!(ds,x)
     ds.t = 1
@@ -173,9 +179,8 @@ end
 # with a random search depending on some distribution. 
 # If the search fails it returns a negative time.
 function get_stagger!(ds, x0, δ, Tm, isinside; max_steps = Int(1e6), f = 1.1, stagger_mode = :exp, verbose = false)
-
     Tp = 0; xp = zeros(length(x0)); k = 1; 
-    T0 = escape_time!(x0, ds, isinside)
+    T0 = escape_time!(ds, x0, isinside)
     if !isinside(x0)
         error("x0 must be in grid")
     end
@@ -190,7 +195,7 @@ function get_stagger!(ds, x0, δ, Tm, isinside; max_steps = Int(1e6), f = 1.1, s
            end
            return 0,-1
         end
-        Tp = escape_time!(xp, ds, isinside)
+        Tp = escape_time!(ds, xp, isinside)
         if stagger_mode == :adaptive
             # We adapt the variance of the search
             # if the alg. can't find a candidate
--- a/src/boundaries/stagger_and_step.jl
+++ b/src/boundaries/stagger_and_step.jl
@@ -1,0 +1,211 @@
+export stagger_and_step!, stagger_trajectory!
+using LinearAlgebra:norm
+using Random
+using ProgressMeter
+
+
+"""
+    stagger_trajectory!(x0 ,ds, isinside; kwargs...) -> xi
+
+This function returns a point `xi` which _guarantees_ `T(xi) > 
+Tm`. This is an auxiliary function for [`stagger_and_step`](@
+ref). Keyword arguments and definitions are identical for both 
+functions. 
+
+The initial search radius is much bigger, `δ = 1.` by default.
+
+"""
+function stagger_trajectory!(x0 ,ds, isinside; δ = 1., Tm = 30, stagger_mode = :exp, max_steps = Int(1e5), f = 1.1)
+    T = escape_time!(x0, ds, isinside)
+    xi = deepcopy(x0) 
+    while T < Tm  # we must have T ≥ Tm at each step 
+        xi, T = get_stagger!(xi, ds, δ, T, isinside; f, stagger_mode, max_steps)
+        if T < 0
+            error("Cannot find a stagger trajectory. Choose a different starting point or search radius δ.")
+        end 
+            
+    end
+    return xi
+end
+
+
+"""
+    stagger_and_step!(x0, ds::DynamicalSystem , N::Int, isinside::function; kwargs...) -> trajectory
+
+This function implements the stagger-and-step method 
+[^Sweet2001] to approximate the invariant non-attracting set 
+governing the chaotic transient dynamics of a system, namely 
+the stable manifold of a chaotic saddle. 
+
+Given the dynamical system `ds` and a initial guess `x0` in a 
+region *with no attractors* defined by the membership function 
+`isinside`, the algorithm provides `N` points close to the 
+stable manifold that escape from the region after a least `Tm` 
+steps of `ds`. The search is stochastic and depends on the 
+parameter `δ` defining a (small) neighborhood of search. 
+
+The function `isinside(x)` returns `true` if the point `x` is 
+inside the chosen bounded region and `false` otherwise. See 
+[`statespace_sampler`](@ref) to construct this function.
+
+## Description 
+
+The method relies on the stagger-and-step algorithm that 
+computes points close to the saddle that escapes in a time 
+`T(x_n) > Tm`. The function `T` represents the escape time 
+from a region defined by the user (see the argument 
+`isinside`).
+
+Given the dynamical mapping `F`, if the iteration `x_{n+1} = 
+F(x_n)` respects the condition `T(x_{n+1}) > Tm` we accept 
+this next point, this is the _step_ part of the method. If 
+not, the method search randomly the next point in a 
+neighborhood following a given probability distribution, this 
+is the _stagger_ part. This part sometimes fails to find a new 
+candidate and a new starting point of the trajectory is chosen 
+within the defined region. See the keyword argument 
+`stagger_mode` for the different available methods.   
+
+The method produces a pseudo-trajectory of `N` points δ-close
+to the stable manifold of the chaotic saddle. 
+
+## Keyword arguments
+* `δ = 1e-10`: it is a small number constraining the random
+  search around a particular point. The interpretation of this 
+  number will depend on the distribution chosen for the 
+  generation (see `stagger_mode`).
+
+* `Tm = 30`: The minimum number of iterations of `ds` before 
+  the trajectory escapes from the bounding box defined by
+  `isinside`. 
+
+* `max_steps = 10^5`: The search for a new candidate point may 
+  fail at some point. If the search fails after `max_steps`, 
+  a new initial point is set and the method starts from a new
+  point 
+
+* `stagger_mode = :exp`: There are several ways to produce 
+  candidate points `x` that have to fulfill the condition 
+  `T(x) > Tm`. The available methods are: 
+
+    * `:exp`: An candidate following an truncated exponential 
+      distribution in a random direction `u` around the current
+      `x` such that `x_c = x + u*r`. `r = 10^-s` with `s` taken
+      from a uniform distribution in [-15, δ]. This mode fails
+      often but stills manage to provide long enough stretch of
+      trajectories. 
+
+    * `:unif`: The next candidate is `x_c = x + u*r` with `r` 
+      taken from a uniform distribution [0,δ]. 
+    * `:adaptive`: The next candidate is `x_c = x + u*r` with 
+      `r` drawn from a gaussian distribution with variance  δ.
+      The variance changes according to a free parameter `f`  
+      such that `δ = δ/f` if no candidate is found and `δ = δ*f`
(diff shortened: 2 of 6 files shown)


Datseris:
Sorry Alex, this is definitely on my todos to review. I am just super overwhelemd in the start of the new term and don't have much time. will review at some point!!!

awage:
No pressure! This is a low priority PR. I did it just to learn a bit more on these techniques.

src/boundaries/example_stagger.jl
Datseris, line 1: You should put this file as a formal example in the Examples documentation page, and perhaps visualize the output. Is there a 2D or 3D example we can use so that the output can be visualized?

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
+export stagger_and_step!, stagger_trajectory!
+using LinearAlgebra:norm
```
Datseris, line 2:
```suggestion
using LinearAlgebra: norm
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+using Random
+using ProgressMeter
+
+
+"""
+    stagger_trajectory!(ds, x0, Tm, isinside; kwargs...) -> xi
+
+This function returns a point `xi` which _guarantees_ `T(xi) >
```
Datseris, line 10:
```suggestion
Return a point `xi` which _guarantees_ `T(xi) > 
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+using ProgressMeter
+
+
+"""
+    stagger_trajectory!(ds, x0, Tm, isinside; kwargs...) -> xi
+
+This function returns a point `xi` which _guarantees_ `T(xi) > 
+Tm` with a random walk search around the initial coordinates 
+`x0`. `T(xi)` is the escape time of the initial condition `xi`.  
+There is no guarantee to find such point, the parameters
```
Datseris, line 13: I am confused here. In the first sentence you said "return a point that guarantees". Now you say there is no guarantee to find such a point...? So what would the function return if it would not find a point?

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+must be adjusted to find the suitable point. This is an .
+auxiliary function for [`stagger_and_step`](@ref). Keyword arguments and 
+definitions are identical for both functions. 
+
+The initial search radius `δ₀` is big, `δ₀ = 1.` by default.
+
+"""
+function stagger_trajectory!(ds, x0, Tm, isinside; δ₀ = 1., stagger_mode = :exp, max_steps = Int(1e5), f = 1.1)
+    T = escape_time!(ds, x0, isinside)
+    xi = deepcopy(x0)
```
Datseris, line 23:
```suggestion
    xi = copy(x0) 
```
why do you needed to deepcopy here? `x0` is at most an array.

src/boundaries/stagger_and_step.jl
Datseris: Ah. I see below that this throws an error instead. Shouldn't error handling be done in the main function though? So that this inner function would return `nothing` instead?

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+            
+    end
+    return xi
+end
+
+
+"""
+    stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; kwargs...) -> trajectory
+
+This function implements the stagger-and-step method
```
Datseris, line 38:
```suggestion
Implement the stagger-and-step method 
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+    end
+    return xi
+end
+
+
+"""
+    stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; kwargs...) -> trajectory
+
+This function implements the stagger-and-step method 
+[^Sweet2001] and [^Sala2016] to approximate the invariant
```
Datseris, line 39: please change citations from `[^Sweet]` to use our BiBTeX integration with `@cite` like in the rest of the library.

src/boundaries/example_stagger.jl
Datseris: The (projected) magnetic pendulum could be used for it as well perhaps?

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+This function returns a point `xi` which _guarantees_ `T(xi) > 
+Tm` with a random walk search around the initial coordinates 
+`x0`. `T(xi)` is the escape time of the initial condition `xi`.  
+There is no guarantee to find such point, the parameters 
+must be adjusted to find the suitable point. This is an .
+auxiliary function for [`stagger_and_step`](@ref). Keyword arguments and 
+definitions are identical for both functions. 
+
+The initial search radius `δ₀` is big, `δ₀ = 1.` by default.
+
```
Datseris, line 19:
```suggestion
The initial search radius `δ₀` is big, `δ₀ = 1.0` by default.
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+This function implements the stagger-and-step method 
+[^Sweet2001] and [^Sala2016] to approximate the invariant 
+non-attracting set governing the chaotic transient dynamics 
+of a system, namely the stable manifold of a chaotic saddle. 
+
+Given the dynamical system `ds` and a initial guess `x0` in a 
+region *with no attractors* defined by the membership function 
+`isinside`, the algorithm provides `N` points close to the 
+stable manifold that escape from the region after a least `Tm`
```
Datseris, line 46:
```suggestion
stable manifold that escape from the region after at least `Tm` 
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+neighborhood following a given probability distribution, this 
+is the _stagger_ part. This part sometimes fails to find a new 
+candidate and a new starting point of the trajectory is chosen 
+within the defined region. See the keyword argument 
+`stagger_mode` for the different available methods.   
+
+The method produces a pseudo-trajectory of `N` points δ-close
+to the stable manifold of the chaotic saddle. 
+
+## Keyword arguments
```
Datseris, line 75: Keyword arguments are _before_ the description section by convention.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+* `δ = 1e-10`: it is a small number constraining the random
```
Datseris, line 76:
```suggestion
* `δ = 1e-10`: A small number constraining the random
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+      The variance changes according to a free parameter `f`  
+      such that `δ = δ/f` if no candidate is found and `δ = δ*f`
+      when it succeeds.    
+* `f = 1.1`: It is the free parameter for the adaptive stagger
+  method. 
+* `δ₀ = 1.`: This is the radius for the first stagger 
+trajectory search. The algorithm looks for a point 
+sufficiently close to the saddle before switching to the 
+stagger-and-step routine. The search radius must be large 
+enough to find a suitable initial.
```
Datseris, line 114:
```suggestion
  trajectory search. The algorithm looks for a point 
  sufficiently close to the saddle before switching to the 
  stagger-and-step routine. The search radius must be large 
  enough to find a suitable initial. 
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+    * `:unif`: The next candidate is `x_c = x + u*r` with `r` 
+      taken from a uniform distribution [0,δ]. 
+    * `:adaptive`: The next candidate is `x_c = x + u*r` with 
+      `r` drawn from a gaussian distribution with variance  δ.
+      The variance changes according to a free parameter `f`  
+      such that `δ = δ/f` if no candidate is found and `δ = δ*f`
+      when it succeeds.    
+* `f = 1.1`: It is the free parameter for the adaptive stagger
+  method. 
+* `δ₀ = 1.`: This is the radius for the first stagger
```
Datseris, line 110:
```suggestion
* `δ₀ = 1.0`: This is the radius for the first stagger 
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+
+    * `:unif`: The next candidate is `x_c = x + u*r` with `r` 
+      taken from a uniform distribution [0,δ]. 
+    * `:adaptive`: The next candidate is `x_c = x + u*r` with 
+      `r` drawn from a gaussian distribution with variance  δ.
+      The variance changes according to a free parameter `f`  
+      such that `δ = δ/f` if no candidate is found and `δ = δ*f`
+      when it succeeds.    
+* `f = 1.1`: It is the free parameter for the adaptive stagger
+  method.
```
Datseris, line 109: You can merge the bullet point of `f` with the above bullet point. Importantly `f` should be changed to some other symbol as `f` is the universal symbol used for the dynamic rule in DynamicalSystems.jl.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+[^Sweet2001]: D. Sweet, *et al.*, Phys. Rev. Lett. **86**, pp 2261  (2001)
+[^Sala2016]: M. Sala, *et al.*, Chaos **26**, pp 123124 (2016)
+"""
+function stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; δ = 1e-10, Tm  = 30, 
+    f = 1.1, max_steps = Int(1e5), stagger_mode = :exp, δ₀ = 1.)
+
+    xi = stagger_trajectory!(ds, x0, Tm, isinside; δ₀, stagger_mode = :unif, max_steps) 
+    v = Vector{Vector{Float64}}(undef,N)
+    v[1] = xi
+@showprogress   for n in 1:N
```
Datseris, line 124: Can you change the `showprogress` to use a `ProgressMeter` and a keyword `show_progress` like we do in the rest of the library?

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+stagger-and-step routine. The search radius must be large 
+enough to find a suitable initial. 
+[^Sweet2001]: D. Sweet, *et al.*, Phys. Rev. Lett. **86**, pp 2261  (2001)
+[^Sala2016]: M. Sala, *et al.*, Chaos **26**, pp 123124 (2016)
+"""
+function stagger_and_step!(ds::DynamicalSystem, x0, N::Int, isinside::Function; δ = 1e-10, Tm  = 30, 
+    f = 1.1, max_steps = Int(1e5), stagger_mode = :exp, δ₀ = 1.)
+
+    xi = stagger_trajectory!(ds, x0, Tm, isinside; δ₀, stagger_mode = :unif, max_steps) 
+    v = Vector{Vector{Float64}}(undef,N)
```
Datseris, line 122: `v` should be replaced with `Vector{typeof(current_state(ds)}(undef, N)`

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+            # The stagger step may fail. We reinitiate the algorithm from a new initial condition.
+            if Tp < 0
+                xp = stagger_trajectory!(ds, x0, Tm, isinside; δ₀, stagger_mode = :exp, max_steps, f) 
+                δ = 0.1
+            end
+            reinit!(ds, xp; t0 = 0)
+        end 
+        step!(ds)
+        xi = deepcopy(current_state(ds))
+        v[n] = xi
```
Datseris, line 138:
```suggestion
        v[n] = copy(current_state(ds)
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+        else
+            xp, Tp = get_stagger!(ds, xi, δ, Tm, isinside; stagger_mode, max_steps, f)
+            # The stagger step may fail. We reinitiate the algorithm from a new initial condition.
+            if Tp < 0
+                xp = stagger_trajectory!(ds, x0, Tm, isinside; δ₀, stagger_mode = :exp, max_steps, f) 
+                δ = 0.1
+            end
+            reinit!(ds, xp; t0 = 0)
+        end 
+        step!(ds)
```
Datseris, line 136: I don't understand why you do this extra step here.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+    return v
+end
+
+
+    
+function escape_time!(ds, x0, isinside) 
+    x = deepcopy(x0) 
+    set_state!(ds,x)
+    reinit!(ds, x; t0 = 0)
+    k = 1; max_step = 10000;
```
Datseris, line 149: `max_step` should be a keyword propagated from the main function

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+    while isinside(x) 
+        k > max_step && break
+        step!(ds)
+        x = current_state(ds)
+        k += 1
+    end
+    return current_time(ds)
+end
+
+function rand_u(δ, n; stagger_mode = :exp)
```
Datseris, line 159: This function can be made significantly more performant by creating dedicated structs that store the dummy variables to reduce allocations, and are then used as function-like-objects. I can do this small change once the PR is ready to go.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+        s = δ*randn()
+        u = (rand(n).- 0.5)
+        u = u/norm(u)
+        return u*s
+    end
+end
+
+# This function searches a new candidate in a neighborhood of x0 
+# with a random search depending on some distribution. 
+# If the search fails it returns a negative time.
```
Datseris, line 181: Instead of having this as comments above the function, make it a documentation string.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,217 @@
...
+function get_stagger!(ds, x0, δ, Tm, isinside; max_steps = Int(1e6), f = 1.1, stagger_mode = :exp, verbose = false)
```
Datseris, line 182: I don't like function names that start with `get_` unless there is a sister name that starts with `set_` or `has_`. Here this isn't the case so this function can be just `stagger!`.

Datseris (commented):
Hi @awage thanks a lot for adding this! I don't have very critical comments at the moment, but I will also give a second review. Two main things I'd like is a plottable example (such as magnetic pendulum, newton or henon maps), and also, what happens in the main function if the max steps of the escape time are exceeded? The trajectory doesn't escape but you also don't do anything about it.

awage:
Thanks for the comments. I will start working this week on this PR.

src/boundaries/stagger_and_step.jl
awage: This is step part of the `stagger_and_step`. If the `escape_time!` is < Tm then  we  execute a step. If not, we look for a neighbor that fulfill the condition and after this we do one step.

awage:
@Datseris I have adressed all your comments. I leave the plotable example in the comments below. It is the coupled Hénon map presented in the publications. If I have the time I will try a continuous system. 

Note: In the test file I implemented a test to verify that the function `escape_time!` throws an error as intended. I am sure that this is the way to test error handling. 


```julia
using Attractors

# Coupled Hénon maps
function F!(du, u ,p, n)
    x,y,u,v = u
    A = 3; B = 0.3; C = 5.; D = 0.3; k = 0.4; 
    du[1] = A - x^2 + B*y + k*(x-u)
    du[2] = x
    du[3] = C - u^2 + D*v + k*(u-x)
    du[4] = u
    return 
end

# The region should not contain any attractors. 
R_min = [-4; -4.; -4.; -4.] 
R_max = [4.; 4.; 4.; 4.]


# Initial box and initial condition.
sampler, isinside = statespace_sampler(HRectangle(R_min,R_max))
x0 = sampler()
df = DeterministicIteratedMap(F!, x0) 
v = stagger_and_step!(df, x0, 10000, isinside; stagger_mode = :adaptive, δ = 1e-4, Tm = 10, max_steps = Int(1e5), δ₀ = 2.) 
v = StateSpaceSet(v)

using CairoMakie
scatter(v[:,1], v[:,3]; markersize = 3)
```

Datseris:
Hi @awage I'm sorry I took so long to come back, life has been very intense for me lately. I still see in this PR that there are no changes in the `docs` folder. At minimum, the documentation string of the main function should be listed somewhere. 

Thanks for attaching an example. My concern is that because this is 4D, how much sense does the visualization make? (can you post the output image please?)

awage:
The projection on 2D of the saddle is like this:

imagen

This is the example published in the original paper.

docs/src/examples.md
```diff
@@ -749,3 +749,64 @@ In this simple two-dimensional model, we could of course have found the saddle d
...
+R_max = [4.; 4.; 4.; 4.]
+sampler, isinside = statespace_sampler(HRectangle(R_min,R_max))
+```
+
+And we are ready! We can now call the function `stagger_and_step` with an initial 
+condition `x0`: 
+
+```@example MAIN
+x0 = sampler()
+v = stagger_and_step!(df, x0, 10000, isinside; stagger_mode = :adaptive, δ = 1e-4, Tm = 10, max_steps = Int(1e5), δ₀ = 2.)
```
awage, line 796:
```suggestion
v = stagger_and_step!(ds, x0, 10000, isinside; stagger_mode = :adaptive, δ = 1e-4, Tm = 10, max_steps = Int(1e5), δ₀ = 2.) 
```

awage:
I have updated the docs. It seems to compile except for a problem with BifurcationToolkit.

Datseris:
Seems like there are undefined references in the docs: https://github.com/JuliaDynamics/Attractors.jl/actions/runs/11835746833/job/32979169108?pr=148#step:5:174  (sorry for the late reply! but next time please check the automated integrations to resolve the errors)

awage:
I completely forgot this PR. I will take some time this week to finish this one. @Datseris Sorry for letting this one opened for so long.

Datseris:
Hi Alex, actually, I am sorry as well. I haven't forgotten about this PR either but I was too busy to do anything. I think once you have the docs I should have a final look at the source code and this should be good to go hopefully!

awage:
@Datseris the PR is passing tests and looks pretty good. I leave it to you.

Datseris:
Perfect, thanks Alex! Coincidentally I am on holiday today so I'll do a proper review!

awage:
Enjoy your holiday! The review can wait.

test/boundaries/stagger_tests.jl
```diff
@@ -0,0 +1,69 @@
...
+
+@testset "Stagger Tests" begin
+
+# Dynamical system with a saddle at 0
+function F!(du, u ,p, n)
+    x,y = u
+    du[1] = x + y 
+    du[2] = x - y
+    return 
+end
```
Datseris, line 12: I love this example. This is brilliant. That's what we need for good tests, analytically resolvable situations!

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,250 @@
...
+        end
+        step!(ds)
+        x = current_state(ds)
+        k += 1
+    end
+    return current_time(ds)
+end
+
+function rand_u(δ, n; stagger_mode = :exp, rng::AbstractRNG)
+    if stagger_mode == :exp
```
Datseris, line 188: In this internal function you should change `stagger_mode` and `rng` into positional arguments. I think the compiler can do constant propagation more robustly and remove the `if` clause this way.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,250 @@
...
+        error("Cannot find a stagger trajectory. Choose a different starting point or 
+                search radius δ₀.")
+    end
+
+    v = Vector{typeof(current_state(ds))}(undef, N)
+    v[1] = xi
+    for n in 1:N
+        show_progress && ProgressMeter.update!(progress, n)
+        if escape_time!(ds, xi, isinside; max_escape_time) > Tm
+            reinit!(ds, xi; t0 = 0)
```
Datseris, line 148: Why is here `t0` imposed as 0? If anything, you shouldn't set `t0` at all, because the default behavior is to set it to the systems _original_ starting time. If you e.g., had a periodically forced dynamical system, this t0 would have meaning. Is there a particular reason to enforce `t0 = 0` here? I don't see it from the docstring.

awage: I don't remember why I set this as 0. I leave the default initial time.

awage:
> Evaluate whether e.g., `@code_warntype` shows you any problems, or if you do a basic profiling something really obvious comes up? Unless you have done this already.

I did a warntype check but I am not sure of about one of the warnings: 

```Julia
 _::Core.Const(Core.kwcall)
  @_2::@NamedTuple{stagger_mode::Symbol, δ::Float64, Tm::Int64, max_steps::Int64, δ₀::Float64}
  @_3::Core.Const(Attractors.stagger_and_step)
  ds::DeterministicIteratedMap{true, Vector{Float64}, 4, typeof(F!), Nothing}
  x0::Vector{Float64}
  N::Int64
  isinside::StateSpaceSets.var"#isinside#61"{Vector{Float64}, Vector{Float64}}
Locals
  δ::Union{}
  Tm::Union{}
  γ::Union{}
  max_steps::Union{}
  max_escape_time::Union{}
  stagger_mode::Union{}
  δ₀::Union{}
  show_progress::Union{}
  rng::Union{}
  @_17::Union{Bool, Float64, Int64, Symbol, Random.Xoshiro}
Body::StateSpaceSet{_A, Float64, SVector{_A1, Float64}} where {_A, _A1}
1 ──       Core.NewvarNode(:(δ))
│          Core.NewvarNode(:(Tm))
│          Core.NewvarNode(:(γ))
│          Core.NewvarNode(:(max_steps))
│          Core.NewvarNode(:(max_escape_time))
│          Core.NewvarNode(:(stagger_mode))
│          Core.NewvarNode(:(δ₀))
│          Core.NewvarNode(:(show_progress))
│          Core.NewvarNode(:(rng))
│          Core.NewvarNode(:(@_17))
│    %11 = Core.isdefined(@_2, :δ)::Core.Const(true)
└───       goto #3 if not %11
2 ──       (@_17 = Core.getfield(@_2, :δ))
└───       goto #4
3 ──       Core.Const(:(@_17 = 1.0e-10))
4 ┄─ %16 = @_17::Float64
│    %17 = Core.isdefined(@_2, :Tm)::Core.Const(true)
└───       goto #6 if not %17
5 ──       (@_17 = Core.getfield(@_2, :Tm))
└───       goto #7
6 ──       Core.Const(:(@_17 = 30))
7 ┄─ %22 = @_17::Int64
│    %23 = Core.isdefined(@_2, :γ)::Core.Const(false)
└───       goto #9 if not %23
8 ──       Core.Const(:(@_17 = Core.getfield(@_2, :γ)))
└───       Core.Const(:(goto %28))
9 ┄─       (@_17 = 1.1)
│    %28 = @_17::Core.Const(1.1)
│    %29 = Core.isdefined(@_2, :max_steps)::Core.Const(true)
└───       goto #11 if not %29
10 ─       (@_17 = Core.getfield(@_2, :max_steps))
└───       goto #12
11 ─       Core.Const(:(@_17 = Attractors.Int64(100000.0)))
12 ┄ %34 = @_17::Int64
│    %35 = Core.isdefined(@_2, :max_escape_time)::Core.Const(false)
└───       goto #14 if not %35
13 ─       Core.Const(:(@_17 = Core.getfield(@_2, :max_escape_time)))
└───       Core.Const(:(goto %40))
14 ┄       (@_17 = 10000)
│    %40 = @_17::Core.Const(10000)
│    %41 = Core.isdefined(@_2, :stagger_mode)::Core.Const(true)
└───       goto #16 if not %41
15 ─       (@_17 = Core.getfield(@_2, :stagger_mode))
└───       goto #17
16 ─       Core.Const(:(@_17 = :exp))
17 ┄ %46 = @_17::Symbol
│    %47 = Core.isdefined(@_2, :δ₀)::Core.Const(true)
└───       goto #19 if not %47
18 ─       (@_17 = Core.getfield(@_2, :δ₀))
└───       goto #20
19 ─       Core.Const(:(@_17 = 1.0))
20 ┄ %52 = @_17::Float64
│    %53 = Core.isdefined(@_2, :show_progress)::Core.Const(false)
└───       goto #22 if not %53
21 ─       Core.Const(:(@_17 = Core.getfield(@_2, :show_progress)))
└───       Core.Const(:(goto %58))
22 ┄       (@_17 = true)
│    %58 = @_17::Core.Const(true)
│    %59 = Core.isdefined(@_2, :rng)::Core.Const(false)
└───       goto #24 if not %59
23 ─       Core.Const(:(Core.getfield(@_2, :rng)))
│          Core.Const(:(%61 isa Attractors.AbstractRNG))
│          Core.Const(:(goto %65 if not %62))
│          Core.Const(:(goto %67))
│          Core.Const(:(%new(Core.TypeError, Symbol("keyword argument"), :rng, Attractors.AbstractRNG, %61)))
│          Core.Const(:(Core.throw(%65)))
│          Core.Const(:(@_17 = %61))
└───       Core.Const(:(goto %70))
24 ┄       (@_17 = Attractors.Xoshiro())
│    %70 = @_17::Random.Xoshiro
│    %71 = Base.keys(@_2)::Core.Const((:stagger_mode, :δ, :Tm, :max_steps, :δ₀))
│    %72 = (:δ, :Tm, :γ, :max_steps, :max_escape_time, :stagger_mode, :δ₀, :show_progress, :rng)::Core.Const((:δ, :Tm, :γ, :max_steps, :max_escape_time, :stagger_mode, :δ₀, :show_progress, :rng))
│    %73 = Base.diff_names(%71, %72)::Core.Const(())
│    %74 = Base.isempty(%73)::Core.Const(true)
└───       goto #26 if not %74
25 ─       goto #27
26 ─       Core.Const(:(Base.kwerr(@_2, @_3, ds, x0, N, isinside)))
27 ┄ %78 = Attractors.:(var"#stagger_and_step#226")::Core.Const(Attractors.var"#stagger_and_step#226")
│    %79 = (%78)(%16, %22, %28, %34, %40, %46, %52, %58, %70, @_3, ds, x0, N, isinside)::StateSpaceSet{_A, Float64, SVector{_A1, Float64}} where {_A, _A1}
└───       return %79
```

The keywords arguments are marked as `Union` and there is one tagged in red. But then it does not give more info about the warning. The warning for StateSpaceSet does not seems a problem.

awage:
> Thanks Alex, this is looking great. Can you please do me a favour and run some standard tests for performance? Evaluate whether e.g., `@code_warntype` shows you any problems, or if you do a basic profiling something really obvious comes up? Unless you have done this already.

I did a profiling with profileview and it seems that everything is ordinary business. I haven't seen any particular function consuming more resources than it should. Most of the work in is `rand_u` for random number generator and for the initialization and iteration of the dynamical system (half of the load for each part).

Datseris (changes requested):
It is unreasonable that the `rand_u` function should take any meaningful amount of time. It is likely because you are testing it with a simple discrete time systems whose time-step takes practically zero time. But this also reveals that the `rand_u` function is suboptimal, and performs a lot of allocations. I've added a commend on how to improve it.

This also reveals that this hasn't been tested with continuous time systems. Is the algorithm generic and works for continuous time systems as well? If so, we need to at least call the function with e.g., the Thomas cyclical (it has fractal basins so it likely has a chaotic saddle) and at least look what the function does there. Can you plot the saddle of Thomas cyclical in 3D and attach the plot? Perhaps based on the plot we can consider a test. Thomas cyclical has this three-fold symmetry maybe we can test for that...?

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,255 @@
...
+        return u*s
+    elseif stagger_mode == :adaptive
+        s = δ*randn(rng)
+        u = randn(rng,n)
+        u = u/norm(u)
+        return u*s
+    else
+        error("Invalid stagger_mode: $stagger_mode")
+    end
+end
```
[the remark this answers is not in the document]
awage: I did the inplace version and there is a 3 fold factor in speed (it goes from 3 to 1 seconds for the Henon map) . I also removed some unnecessary allocation.

docs/src/api.md
```diff
@@ -98,13 +98,15 @@ uncertainty_exponent
...
-The edge tracking algorithm allows to locate and construct so-called edge states (also referred to as *Melancholia states*) embedded in the basin boundary separating different basins of attraction. These could be saddle points, unstable periodic orbits or chaotic saddles. The general idea is that these sets can be found because they act as attractors when restricting to the basin boundary.
+## Edge tracking, edge states and chaotic saddles
+The edge tracking algorithm allows to locate and construct so-called edge states (also referred to as *Melancholia states*) embedded in the basin boundary separating different basins of attraction. These could be saddle points, unstable periodic orbits or chaotic saddles. The general idea is that these sets can be found because they act as attractors when restricting to the basin boundary. Another technique to get a pseudo trajectory close to a saddle is the stagger-and-step method that requires little information on the dynamical system. 
 
 ```@docs
 edgetracking
 EdgeTrackingResults
 bisect_to_edge
+stagger_and_step
+stagger_trajectory
```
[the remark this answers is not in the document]
awage: It is not a necessary function for the API. I remove this and the export also.

awage:
> Can you plot the saddle of Thomas cyclical in 3D and attach the plot? Perhaps based on the plot we can consider a test. Thomas cyclical has this three-fold symmetry maybe we can test for that...?

I tried the Hénon Heiles system but I had a lot of issues with kam Islands and sticky trajectories. I will look for the Thomas 3D or any other continuous system

awage:
I made a crude test with the Thomas 3D and it finds the saddle at the origin. The profiling now says that 95-ish%  of the work is the integration and reinitialization stuff.

src/boundaries/stagger_and_step.jl
```diff
@@ -173,39 +173,37 @@ end
...
+        u .= randn(rng,n)
+        u .= s*u/norm(u)
+        return
     elseif stagger_mode == :adaptive
         s = δ*randn(rng)
-        u = randn(rng,n)
-        u = u/norm(u)
-        return u*s
+        u .= randn(rng,n)
+        u .= s*u/norm(u)
```
Datseris, line 205: `s*u` still creates a new vector. Why are you not doing simply `u .*= s/norm(u)`?

awage: I still have a lot to learn in optimizing Julia code. I assumed that the compiler took care of these kind of memory allocations.

Datseris:
> I have checked the Thomas oscillator and the algorithm finds (a part of) the stable manifold of the saddle. This is working. I haven't computed the eigenvalues of the fixed point at (0,0,0) but it seems a complex eigevalues with positive real part.

Yes, this is correct. We can deduce this even from the symmetry of the system. Can you please add a test with the thomas model? just test that the saddle is approximately the 0 vector.

src/boundaries/stagger_and_step.jl
```diff
@@ -192,17 +192,17 @@ function rand_u!(u, δ, n, stagger_mode, rng)
         a = -log10(δ)
         s = (15-a)*rand(rng) + a
         u .= randn(rng,n)
-        u .= 10.0^-s*u/norm(u)
+        u .*= 10.0^-s/norm(u)
```
Datseris, line 195:
```suggestion
        u .*= (10.0^(-s))/norm(u)
```

for readability. I can never tell what operation happens first and I don't remember by heart the order of operations.

test/boundaries/stagger_tests.jl
```diff
@@ -1,5 +1,6 @@
-using Attractors
+using Attractors, OrdinaryDiffEq
```
Datseris, line 1: You are not using OrdinaryDiffEq anywhere in this file.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,255 @@
+export stagger_and_step
+using LinearAlgebra: norm
+using Random
+using ProgressMeter
+
+
+
+
```
Datseris, line 8:
```suggestion
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,255 @@
...
+      when it succeeds. 
+
+* `γ = 1.1`: It is free parameter for the adaptive stagger method `:adaptive`. 
+
+* `δ₀ = 1.0`: This is the radius for the first stagger 
+  trajectory search. The algorithm looks for a point 
+  sufficiently close to the saddle before switching to the 
+  stagger-and-step routine. The search radius must be large 
+  enough to find a suitable initial candidate. 
+
```
Datseris, line 79:
```suggestion
  enough to find a suitable initial candidate. 
  To type `δ₀` use `\delta<TAB>\_0<TAB>`.

```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,255 @@
...
+* `rng::AbstractRNG = Xoshiro()`:  Random number generator. Use this for
+  reproducibility.
+
+## Description 
+
+The method relies on the stagger-and-step algorithm that 
+search initial conditions close to the saddle with escapes time
```
Datseris, line 86:
```suggestion
searches initial conditions close to the saddle with escapes time 
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,255 @@
...
+
+## Keyword arguments
+* `δ = 1e-10`: A small number constraining the random
+  search around a particular point. The interpretation of this 
+  number will depend on the probability distribution chosen 
+  for the sampling (see `stagger_mode`).
+
+* `Tm = 30`: The minimum number of iterations of `ds` before 
+  the trajectory escapes from the bounding box defined by
+  `isinside`.
```
Datseris, line 41: With respect to `T` and `Tm`: what do these mean for continuous time systems? In DynamicalSystems.jl we should try to have a consistent terminology. Let's not use the "iteration" word. Let's just use "steps". This directly refers to the `step!` function. Since this algorithm applies to continuous time systems as well, this function should also allow for the keywords `Δt` and `stop_at_Δt`, which are propagated to `step!`. Just like most other functions of the library.

Datseris: This also means that you should change all `step!` calls to `step!(ds, Δt, stop_at_Δt)`. By default `Δt = 1` and `stop = false`.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,255 @@
...
+`x0`. `T(xi)` is the escape time of the initial condition `xi` from 
+the bounding box defined by the function `isinside`.  
+In the case where the algorithm cannot find a suitable point, the algorithm 
+returns nothing. 
+
+This is an auxiliary function for [`stagger_and_step`](@ref). 
+Keyword arguments and definitions are identical for both functions. 
+"""
+function stagger_trajectory(ds, x0, Tm, isinside; δ₀ = 1., stagger_mode = :exp,
+        max_steps = Int(1e5), γ = 1.1, max_escape_time = 10000, rng::AbstractRNG = Xoshiro())
```
Datseris, line 160: Defaults should only be given at one place. For easier maintainance, you shouldn't specify `max_escape_time = 10000` in three functions. You should only specify it at the top-level function. The rest of the functions should either have `max_escape_time` as a positiononal arguemnt, or in this case, as a mandatory keyword argument. This goes for all other keywords.

test/boundaries/stagger_tests.jl
```diff
@@ -0,0 +1,89 @@
...
+for u in v
+    @test Attractors.escape_time!(df, u, isinside) ≥ 30
+end
+
+# Test if all the points have escape time ≥ Tm 
+# :adaptive mode
+v = stagger_and_step(df, xi, 10, isinside; δ₀ = 1e-3, stagger_mode = :adaptive) 
+for u in v
+    @test Attractors.escape_time!(df, u, isinside) ≥ 30 - 1
+end
```
Datseris, line 46: You can create a `for` loop with the three stagger modes to make the code cleaner.

test/boundaries/stagger_tests.jl
```diff
@@ -0,0 +1,89 @@
...
+df = DeterministicIteratedMap(F_stable!, x0) 
+# Test if the program throws an error when the 
+# trajectory stays inside the square
+try 
+    u = sampler()
+    Attractors.escape_time!(df, u, isinside)
+    @test false
+catch 
+    @test true  
+end
```
Datseris, line 65: There is a better way to test for errors. The current setup doesn't test for what error you are getting. You should instead use the `@test_throws` macro and remove this `try`-block alltogether.

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,255 @@
...
+function rand_u!(u, δ, n, stagger_mode, rng)
+    if stagger_mode == :exp 
+        a = -log10(δ)
+        s = (15-a)*rand(rng) + a
+        u .= randn(rng,n)
+        u .*= (10.0^(-s))/norm(u)
+        return
+    elseif stagger_mode == :unif
+        s = δ*rand(rng)
+        u .= randn(rng,n)
```
Datseris, line 199:
```suggestion
        randn!(rng, u)
```

this still allocates!!! You have to use in-place operations. Please read the documentation strings of `randn` and `randn!`. You need to do `using Random: randn!`.

src/boundaries/stagger_and_step.jl
Datseris: Or, do you think it is better for continuous time systems to only use the integrator iteration? So if the actual amount of time the system evolves doesn't matter, then ignore my suggestion. But still simply change "iteration" to "step".

Datseris: Or, do you think it is better for continuous time systems to only use the integrator iteration? So if the actual amount of time the system evolves doesn't matter, then ignore my suggestion. But still simply change "iteration" to "step".

Project.toml
```diff
@@ -2,7 +2,9 @@ name = "Attractors"
 uuid = "f3fd9213-ca85-4dba-9dfd-7fc91308fec7"
 authors = ["George Datseris <datseris.george@gmail.com>", "Kalel Rossi", "Alexandre Wagemakers"]
 repo = "https://github.com/JuliaDynamics/Attractors.jl.git"
-version = "1.24.5"
+
+version = "1.25"
+
```
Datseris, line 7:
```suggestion
version = "1.25"
```

docs/src/examples.md
```diff
@@ -749,3 +748,64 @@ In this simple two-dimensional model, we could of course have found the saddle d
...
+The mode `:adaptive` adapts the radius of the stochastic
+search as a function of the success of the search process. 
+
+Finally we can represent a projection of the chaotic saddle found in this 
+example: 
+
+```@example MAIN
+fig = Figure()
+ax = Axis(fig[1,1], xlabel="x", ylabel="y")
+v = StateSpaceSet(v)
```
Datseris, line 808:
```suggestion
```

src/boundaries/stagger_and_step.jl
```diff
@@ -0,0 +1,259 @@
...
+    k = 1; 
+    while isinside(current_state(ds)) 
+        if k > max_escape_time 
+            error("The trajectory did not escape for ", k, " steps, you probably have \
+an attractor in the defined region. Last point evaluated: ", current_state(ds))
+        end
+        step!(ds)
+        k += 1
+    end
+    return current_time(ds)
```
Datseris, line 189:
```suggestion
    return current_time(ds) - initial_time(ds)
```

This is correct for systems whose initial time happens to not be 0

Datseris (approved):
I'll let the test and doc build test. If they pass I'll merge! If not, can you please ensure they do?

PR #149, opened 2024-09-18
Small improvements to notebook conversion and GroupViaPairwiseComparison

Datseris:


Datseris:
(i am merging this in without review because it didn't really do any meanignful changes. Besides adding a default value for `threshold` that makes some sense if rescaling is used).

PR #160, opened 2024-10-16
plot continuation using scatterlines and NaNs

Datseris:
this makes it more efficient as well, coincidentally. Not that it matters because it is unikely that we plot millions of points during the continuation.


Datseris:
Looks good I think: 

f2d6ef40

Issue #159, opened 2024-10-15
Type error when running `convergence_and_basins_of_attraction`


andreasmorr:
**Describe the bug**
When running `convergence_and_basins_of_attraction` with any mapper, I get an `Inexact error: Int64`. I believe this is due to the array `iterations` within `convergence_and_basins_of_attraction` being initialized with type Int while it is intended to store Float times. Also, `basins` should be initialized with type Int so there are no issues when using the entries as keys to `attractors` right?

**Minimal Working Example**

```julia
using Pkg; Pkg.add("DynamicalSystems")
Pkg.add("Attractors")
Pkg.instantiate()

using DynamicalSystems
using Attractors

function fold_nf_rule!(du, u, p, t)
    du[1] = -u[1]^2 + p[1]
    return nothing
end

u0 = [10.0]
p0 = [1.0]

fold_nf_ds = CoupledODEs(fold_nf_rule!, u0, p0)

grid = (range(-2, 2; length = 41),)

mapper = AttractorsViaRecurrences(fold_nf_ds, grid; sparse = false)

basins, attractors, convergence = convergence_and_basins_of_attraction(mapper, grid; show_progress = true)
```

**Package versions**

```julia
Status `~/.julia/environments/v1.10/Manifest.toml`
  [f3fd9213] Attractors v1.21.0
  [61744808] DynamicalSystems v3.3.22
```

awage:
Hi! I will have a look at this.

awage:
Looks like a bug. I will fix it.

andreasmorr:
Thanks!

PR #161, opened 2024-10-16
Fix convergence time for continuous dynamical systems

awage:
The function `convergence_and_basins_of_attraction` throws an error when the dynamical is a continuous dynamical system. 

See issue #159

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "Attractors"
 uuid = "f3fd9213-ca85-4dba-9dfd-7fc91308fec7"
 authors = ["George Datseris <datseris.george@gmail.com>", "Kalel Rossi", "Alexandre Wagemakers"]
 repo = "https://github.com/JuliaDynamics/Attractors.jl.git"
-version = "1.22.1"
+version = "1.22.0"
 
 
 [deps]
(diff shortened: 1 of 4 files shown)


src/basins/basins_utilities.jl
```diff
@@ -86,7 +86,11 @@ function convergence_and_basins_of_attraction(mapper::AttractorMapper, grid; sho
         @error "The mapper and the grid must have the same dimension"
     end
     basins = zeros(length.(grid))
-    iterations = zeros(Int, length.(grid))
+    if isa(mapper.ds,ContinuousDynamicalSystem)
+        iterations = zeros(length.(grid))
+    else
+        iterations = zeros(Int, length.(grid))
+    end
```
Datseris, line 93: Instead of doing all of this we can do `zeros(typeof(current_time(ds)), ...)`. No need for Ifs.

Also, please don't use `mapper.ds`. Use instead `referenced_dynamical_system(mapper)`.

Project.toml
```diff
@@ -2,7 +2,8 @@ name = "Attractors"
 uuid = "f3fd9213-ca85-4dba-9dfd-7fc91308fec7"
 authors = ["George Datseris <datseris.george@gmail.com>", "Kalel Rossi", "Alexandre Wagemakers"]
 repo = "https://github.com/JuliaDynamics/Attractors.jl.git"
-version = "1.22.0"
+version = "1.22.1"
```
Datseris, line 5:
```suggestion
version = "1.22.0"
```

Datseris (approved):

Datseris:
Ah wait, this fixes #159 right? We should add a simple test in the test suite that checks for this. Its good practice when we fix a bug to add a test that checks that it was fixed for future proofness. Can you please use #159 to turn it into a simple unit test?

awage:
@Datseris  the test is ready.

Datseris (approved):

PR #165, opened 2025-01-21
[WIP] nonlocal stability measure estimator

Datseris:
This PR is by @andreasmorr and adds a nonlocal stability estimator construct that can compute every nonlocal stability measure at once during finding attractors and/or during the continuation.


src/mapping/stability_measures_accumulator.jl
```diff
@@ -0,0 +1,351 @@
...
+* `basin_fractions`: The fraction of initial conditions that converge to the
+  attractor.
+* `finite_time_basin_fractions`: The fraction of initial conditions that
+  converge to the attractor within the time horizon `T`.
+* `basin_stability`: Same as `basin_fractions`, but the initial conditions are
+  weighted by the probability density of the distribution `d`.
+* `finite_time_basin_stability`: Same as `finite_time_basin_fractions`, but the
+  initial conditions are weighted by the probability density of the distribution
+  `d`.
+* `persistence`: Trajectories from all points of the attractor are evolved under
```
Datseris, line 100: I think this is special. Measures that require parameter change do not fit to this interface. We need to create another function for them I think...

src/mapping/stability_measures_accumulator.jl
```diff
@@ -0,0 +1,351 @@
...
+  the alternative parameter setting `p`. The persistence is the time at which
+  one of the trajectories first leaves the original basin of attraction of the
+  attractor. If the trajectories do not leave the basin of attraction, the
+  persistence is set to `Inf`.
+"""
+mutable struct StabilityMeasuresAccumulator <: AttractorMapper
+    mapper::AttractorMapper
+    ds::DynamicalSystem
+    basin_points::Dict{Int64, StateSpaceSet}
```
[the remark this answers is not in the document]
andreasmorr: What's the cleanest way to extract the specific `{D, T}` from a given `mapper`?

src/mapping/stability_measures_accumulator.jl
```diff
@@ -0,0 +1,351 @@
...
+    finite_time_basin_points::Dict{Int64, StateSpaceSet}
+    nonzero_measure_basin_points::Dict{Int64, StateSpaceSet}
```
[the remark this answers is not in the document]
andreasmorr: I think the stored points are different no? You mean line 110 and 111? One includes the points that converge after time T and the other stores those where d has nonzero measure/pdf.

src/mapping/stability_measures_accumulator.jl
andreasmorr: Yes I agree. Can we still include it in this PR/ file though?

src/mapping/stability_measures_accumulator.jl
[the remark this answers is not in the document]
Datseris:
```
D = dimension(ds)
T = eltype(current_state(ds))
```

src/mapping/stability_measures_accumulator.jl
Datseris: Sure, if you want to, but it has to be a separate function for source code clarity. It would be a function you run after `global_continuation` has finished and you utilize the `attractors_cont` output.

Datseris:
plotting is simple, we don't need a specialized function:
```julia
attractors_cont, all_measures = global_cont(stability_accumulator)

all_measures = vector ( dictonary ( measure_name => dictionary (id => number) ) )

mcs = [dicts["minimal_fatal_shock_magnitude"] for dicts in all_measures]

plot_continuation_curves(mcs, parameter_curve)
```

Datseris:
I just found the paper: https://journals.aps.org/pre/abstract/10.1103/PhysRevE.93.042205 "Constrained basin stability for studying transient phenomena in dynamical systems".

It is identical to estimating the basin stability, but instead of weighting the distribution by a given density ρ, you weight the distribution by a "selector function". Only trajectories whose transients satisfy property "X" are counted. I think it is very straightforward to add this as yet another indicator in this PR.

We make a function `indicator(u0, ds, tc)` with `ds` the dynamical system. Inside this function the user may perform arbitrary numerics of whether the trajectory starting from u0, with arbitrary transient, satisfies the condition. `tc` is the convergence time that we have already estimated from the given initial condition at the previous step of the aglroithm of estimating all other stability measures. then the `indicator` function returns true/false, which tells whether to add `u0` to the count or not.

@andreasmorr I can add this functionality in the code once the PR is ready to go.

src/mapping/stability_measures_accumulator.jl
```diff
@@ -0,0 +1,328 @@
...
+    mapper::AM
+    basin_points::Dict{Int64, StateSpaceSet{Dims, Datatype}}
+    finite_time_basin_points::Dict{Int64, StateSpaceSet{Dims, Datatype}}
+    nonzero_measure_basin_points::Dict{Int64, StateSpaceSet{Dims, Datatype}}
+    mean_convergence_time::Dict{Int64, Float64}
+    maximal_convergence_time::Dict{Int64, Float64}
+    mean_convergence_pace::Dict{Int64, Float64}
+    maximal_convergence_pace::Dict{Int64, Float64}
+    # TODO: It is wrong to say that T is a float. it should be type parameterized.
+    T::Float64
```
Datseris, line 110: Remember this todo, `T` should be an intependent type parameter because it is float for continuous time but integer for discrete time systems.

src/continuation/continuation_stability_measures.jl
```diff
@@ -0,0 +1,81 @@
...
+        for key in attractors_cont_keys
+            ps_unstable = [prange[j] for j in 1:length(prange) if !haskey(char_ret_time_cont[j], key) || (char_ret_time_cont[j][key]==Inf64 || isnan(char_ret_time_cont[j][key]))]
+            if haskey(char_ret_time_cont[i], key) && length(ps_unstable)>0 && (char_ret_time_cont[i][key]!=Inf64 && !isnan(char_ret_time_cont[i][key]))
+                distance_to_bifurcation[end][key] = minimum([abs(ps_unstable[k] - prange[i]) for k in 1:length(ps_unstable)])
+            else
+                distance_to_bifurcation[end][key] = NaN
+            end
+        end
+    end
+    merge!(measures_cont, Dict("distance_to_bifurcation" => distance_to_bifurcation))
```
Datseris, line 79: You also need to remove this distance to bifurcation here; similarly to the persistence this measure does not fit the design. It is completely fine for it to be a different function.

Datseris:
@andreasmorr have a look at my latest commit please. I've improved a bit the interface to use the existing call signature of `global_continuation` with `AttractorSeedContinueMatch`. No need to have a special method.

When using this in the project, just set the `seeding` keyword for ACAM. Don't make a new function that has a different default instead.

I've also removed the distance to bifurcation function as it does not fit at all the generallity of `global_continuation`. Moreover, there is no reason to assume only a single bifurcation occurs during the global continuation. In the examples I am using these methods, there are several global bifurcations happening.

Lastly, don't forget that this functionality needs tests before it is merged. Something needs to be added in the `test` folder. It also needs documentation: a simple example that is added in the Examples page of the documentation.

ext/plotting.jl
```diff
@@ -295,15 +295,15 @@ function Attractors.plot_continuation_curves!(ax, continuation_info, prange = 1:
         labels = Dict(ukeys .=> ukeys),
         add_legend = length(ukeys) < 7,
         markers = markers_from_keys(ukeys),
-        slines_kwargs = (;linewidth = 3,),
+        series_kwargs = NamedTuple(),
```
Datseris, line 298: Why has this changed?

src/continuation/continuation_stability_measures.jl
```diff
@@ -26,7 +26,7 @@ function global_continuation(
     ProgressMeter.next!(progress)
 
     # Continue loop over all remaining parameters
-    ### For George: Is this what you had in mind when we emailed about 
+    ### For @Datseris: Is this what you had in mind when we emailed about
```
andreasmorr, line 29: @Datseris Here I try to set up an adhoc continuation of a StabilityMeasuresAccumulator that has a mapper of type AttractorsViaProximity. In each continuation step, I define a dummy mapper based on Recurrences, that should then continue the attractors one step before calculating the stability measures based on the Proximity mapper again. The problem right now is, that without explicitly passing the grid, we can seem to define an adhoc AttractorsViaRecurrences instance. What do you think?

Datseris: > I define a dummy mapper based on Recurrences, that should then continue the attractors one step before calculating the stability measures based on the Proximity mapper again.

What do you mean "again"? Isn't the starting mapper a recurrences mapper anyways? I also do not see anywhere in the source code the creation of a proximity mapper. Where are you referring to?

---

Line 44 `dummy_mapper = AttractorsViaRecurrences(referenced_dynamical_system(mapper), ics)` fails, and this is expected. Have a look at the docstring for `AttractorsViaRecurrences`. The second input has to be a grid. The second input to AVR is NOT an initial conditions sampler. It is a state space tesselation. 

In general, attractor mappers **have nothing to do with initial conditions sampling**. The other downstream functions utilize the random sampling. E.g., If you use `basins_fractions`, there the second input argument is an initial conditions sampler.

Does this make sense now?

src/continuation/continuation_stability_measures.jl
```diff
@@ -67,43 +53,61 @@ function global_continuation(
...
-            push!(transposed_measures_cont[measure_name], measure_dict)
+            push!(transposed[measure_name], measure_dict)
         end
     end
-
-    # And lastly, we also match sequentially all dictionaries here
-    for measure_cont in values(measures_cont)
-        match_sequentially!(measure_cont, rmaps)
-    end
+    measures_cont = transposed
```
Datseris, line 73: unecessary. Just retun `transposed`. You are changing the type of `measures_cont` which is in general bad practice in Julia.

src/continuation/continuation_stability_measures.jl
```diff
@@ -67,43 +53,61 @@ function global_continuation(
...
+    for (i, p) in enumerate(pcurve)
+        ε = εs isa AbstractVector ? εs[i] : εs # if its a vector, get i-th entry
+        d = distributions isa AbstractVector ? distributions[i] : distributions # if its a vector, get i-th entry
+        T = Ts isa AbstractVector ? Ts[i] : Ts # if its a vector, get i-th entry
+        set_parameters!(ds, p)
+        attractors = attractors_cont[i]
+        accumulator = StabilityMeasuresAccumulator(
+            AttractorsViaProximity(ds, attractors, ε, proximity_mapper_options...),
+            d=d, T=T, metric=metric
+        )
```
Datseris, line 93:
```suggestion
            AttractorsViaProximity(ds, attractors, ε; proximity_mapper_options...),
            d=d, T=T, metric=metric
        )
```

ext/plotting.jl
andreasmorr: It was like this when I started. It's still like this in the main branch.

andreasmorr: It was like this when I started. It's still like this in the main branch.

src/mapping/stability_measures_accumulator.jl
andreasmorr: It's out of the PR now.

src/continuation/continuation_stability_measures.jl
```diff
@@ -51,7 +49,8 @@ function global_continuation(
         push!(measures_cont, measures)
     end
 
-    rmaps = match_sequentially!(attractors_cont, acam.matcher; pcurve, ds = referenced_dynamical_system(mapper))
+    rmaps = match_sequentially!(attractors_cont, acam.matcher; pcurve,
+                                ds = referenced_dynamical_system(mapper))
```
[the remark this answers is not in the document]
andreasmorr: Ah I see! I've applied this wherever this solves the length problem. How are lines with list comprehension etc usually handled? For example, how would you format this line:
`minimal_fatal_shock_magnitudes[key1] = minimum([set_distance(attractors[key1], accumulator.nonzero_measure_basin_points[key2], StateSpaceSets.StrictlyMinimumDistance(true, accumulator.metric)) for key2 in keys(accumulator.nonzero_measure_basin_points) if key1 != key2])`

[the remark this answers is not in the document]
andreasmorr: Or if else statements like this
`basin_stab = Dict(k => (isempty(v) ? 0.0 : sum(pdf(accumulator.weighting_distribution, point) for point in v)) for (k, v) in bpoints)`

Datseris:
I don't understand. Indeed, none of the edge tracking tests pass if the `AttractorsViaProximity` mapper does not obtain the keyword `Ttr = 100`. I don't understand why. The behavior of the proximity mapper should not depend fundamentally on Ttr, provided that lost step counter is sufficiently high.

Datseris:
@andreasmorr I am very worried. It appears that the bahavior of `AttractorsViaProximity` depends fundamentally on whether `Ttr` is given or not? I don't understand. It really shouldn't. Both edge tracking and minimal fatal shock fail without `Ttr = 100`. Have a look at the file `tests/mfs/mfstests.jl` at line 134:
```julia
mapper_m = AttractorsViaProximity(psys, attractors_m; Ttr = 100)
```
this is the mapper used to find the MFS for the magnetic pendulum. If I don't give `Ttr = 100` the tests fail. Initially I thought it was because we didn't allow for enough iterations to reach the attractor, because now we dont have any transient time, and we start the "lost counter" immediatelly. But no, because even if I set
```julia
mapper_m = AttractorsViaProximity(psys, attractors_m; consecutive_lost_steps = 10000000000)
```

the MFS test still fails meaning we don't find the correct shock. What is going on here???? I am really at a loss.

Datseris:
please please please have a look and a deep think on why this is happening. It is really critical. We really must be able to answer why this is happening since we rely so much on the proximity mapper also for our new feature (stability continuation).

Datseris:
@andreasmorr tests fail. https://github.com/JuliaDynamics/Attractors.jl/actions/runs/14495559478/job/40662271289?pr=165#step:5:978 . When you rename something you have to make sure you rename it everywhere. 
```
│    KeyError: key "minimal_fatal_shock_magnitude" not found`
```
because docs try to access this key.

Datseris:
@andreasmorr there is a problem. Have a look at this code:
```
@testset "Henon map linear measures" begin
    henon_rule_alter(x, p, n) = SVector{2}(1.0 - p[1]*x[1]^2 + x[2], -p[2]*x[1])
    μ = 1.05; J = 0.9
    ds = DeterministicIteratedMap(henon_rule_alter, zeros(2), [μ, J])
    xg = range(-3.0, 3.0; length = 101)
    yg = range(-3.0, 4.0; length = 101)
    grid = (xg, yg)


    mapper = AttractorsViaRecurrences(ds, grid; sparse = false, consecutive_recurrences = 1000)
    accumulator = StabilityMeasuresAccumulator(mapper)


    A = ics_from_grid(grid)
    for u0 in A
        id = accumulator(u0)
    end
    stability_measures = finalize_accumulator(accumulator)

    linear_measures = ["characteristic_return_time", "reactivity", "maximal_amplification", "maximal_amplification_time",]

    @testset "Henon $m" for m in linear_measures
        measures = collect(values(stability_measures[m]))
        @test count(!isnan, measures) ≥ 1
    end
end
```

This dynamical system has two attractors: a period 1 (fixed point) and a period 3. Shouldn't all linear stability measures be calculated for hte fixed point? When you run this test you will notice that _all_ linear stability measures are `NaN`. Can you please explain?

Datseris:
Hm.... I don't know. Aren't all linear stability measures straightforward analytic functions of the jacobian eigenvalues? Fixed points have jacobian eigenvalues irrespectively of if the dynamical system is discrete or continuous time... Its just that in one case the jacobian eigenvalue is the exponential rate, while in the discrete time case the jacobian eigenvalue is the direct rate.

We can make the decision to say that "for x/y reason" it doesn't make sense to estimate local stability for discrete systems? Unfortunatly, I cant' imagine such an x/y reason.

andreasmorr:
@Datseris Ah interesting, I have never heard of the direct rate. Do you have a reference?

My view would be that "linear" always refers to a linear relationship between the distance to the attractor and the restoring rate. This is well defined for smooth continuous-time systems because there you can write the linear relationship by Tayloring around the attractor. For discrete-time systems you could in principle have an attractor that's isolated in space, so that all ics in a ball around the attractor converge to something else. A linear relationship of distance to attractor and restoring rate is not possible.

andreasmorr:
Or asked another way, what is the Jacobian of a discrete time system?

Datseris:
In my view "linear" just means linearized, the stability indicators that are based on the linearized system. The Jacobian of a discrete system is the same as that of a continuous one! I define this in detail, including the rates of increase of a pertubation, in chapter 1, section 1.4 here:  https://link.springer.com/book/10.1007/978-3-030-91032-7

andreasmorr:
Ah I see! Ok then let's include that. We need to simply remove` if !isdiscretetime(ds)` in line 326 of stability_measures_accumulator.jl. However, the tests will fail then because the dumb_map is very degenerate and forward diff can't deal with that (?). Do you have an idea how to try/catch the computation of Jacobian in lines 336/337 so that such cases are excluded?

Datseris:
when you talk about lines of code please say which file you are referring to...

andreasmorr:
> ... stability_measures_accumulator.jl ...

Datseris:
It is not so simple @andreasmorr . You have to take into account the nature of discrete time and re-write the necessary expressions for linear measures. One more downside of linear measures, in contrast to nonlinear like the return time or the MCS, is that they depend on the nature of time. The expression you have for `maximal_amplification` is only valid for continuous time... 

I will merge this in now saying that discrete time measures are not computed, but please consider opening a second Pull Request in the future that implements the linear measures for discrete time. My textbook provides the correct expressions for the linear time evolution in discrete time.

PR #168, opened 2025-03-10
Fix BifKit erroring in docs

Datseris:


rveltz:
Hi,

I just looked at this. First the defaults of collocation are not great for speed. It can go down to 0.3s in your example. Remove most allocations with `periodic_orbit_algo = BK.PeriodicOrbitOCollProblem(40, 4; jacobian = BK.DenseAnalyticalInplace())`

Also, you dont get the branch before p=5 because you continue from p = 5 to larger values. Do the following if you want to have bothside. But then, you are limited by `p_min` ; you can adjust to smaller values if you want.

```julia
@time branch = BK.continuation(probpo, cish, predictor, opts_br;
    verbosity = 0, plot = false,
    linear_algo = BK.COPBLS(), # faster linear solver
    argspo...,
    bothside = true,
)
```

rveltz:
Oh and `abs(z)` in the vector field. I am even surprised that it works given it is not differentiable...

The above options for collocation will become the default soon, I am battle testing this

Datseris:
Thanks so much, I was very frustrated that it stopped working, but I didn't want to ask at BFKit to not waste your time... I will try your suggestions in another PR.

Regarding the `abs(z)`: is it really such a big problem? it is only a single plane in 3D space where it is not differentiable. I guess you have a point, but the derivative is ForwardDiff, so it somehow takes the `abs(z)` into account?

Datseris:
Also, why is it that the suggestions you have here for faster Jacobian and linear algorithm are not the default options?

Datseris:
> Also, you dont get the branch before p=5 because you continue from p = 5 to larger values

I don't understand this comment. I have
```
opts_br = BK.ContinuationPar(
    p_min = prange[1], p_max = prange[end],
```
and `prange[1]` is 4.7. Why do I start from p = 5?

Datseris:
Great, your suggestions make it work. They also take 2 seconds (on my computer). The picture is:

image

(still, there is no periodic orbit below 4.8, so perhaps the comments I have in the text do make sense: it could be a quasi periodic or very weakly chaotic)

They do throw a warning:
```
┌ Warning: Error in eigenvalues computation. It seems that an eigenvalue is missing, probably conj(λ) for some already computed eigenvalue λ. This makes the identification (but not the detection) of bifurcation points erroneous. You should increase the number of requested eigenvalues `nev`.└ @ BifurcationKit C:\Users\datse\.julia\packages\BifurcationKit\3l0Yy\src\Bifurcations.jl:113   
  2.078911 seconds (2.08 M allocations: 703.727 MiB, 3.05% gc time)
```
Could this be related with the discontinuous derivative?

rveltz:
> but I didn't want to ask at BFKit to not waste your time

You should! open an issue or ping me

rveltz:
> Why do I start from p = 5?

It starts from the current value you pass in `BifurcationProblem`

rveltz:
You are making a mistake: the period is 5 initially, not 19

Also, it seems to converge to a homoclinic orbit

Screenshot 2025-03-11 at 11 23 02 PM

rveltz:
Screenshot 2025-03-11 at 11 26 33 PM

rveltz:
It is a complicated example, this is what I used (should be improved)

```julia

# BK requires to modify the rule with t=0 as well
function modified_lorenz_rule!(du, u, p, t = 0)
    x, y, z = u; a, b = p
    du[1] = y - x
    du[2] = - x*z + b*abs(z)
    du[3] = x*y - a
    return du
end
prange = 4.7:0.02:6
pidx = 1
p0 = [5.0, 0.1] # parameters
u0 = [-4.0, 5, 0] # state

import BifurcationKit as BK
using OrdinaryDiffEq
using GLMakie
Makie.inline!(true)

bf_prob = BK.BifurcationProblem(
    modified_lorenz_rule!, u0, p0, (BK.@optic _[pidx])
)

# This guess comes from the Attractors.jl main tutorial
point_on_lc = [
    -1.622956992666447,
    -4.527917984019188,
    -5.178825669659272,
]

ode_prob = ODEProblem(modified_lorenz_rule!, point_on_lc, (0.0, 100.0), p0)
sol = OrdinaryDiffEq.solve(ode_prob; alg = Vern9(), abstol = 1e-9, reltol = 1e-9)
j = length(sol)÷2
fig, ax = lines(sol.t[j:end], sol[1, j:end])
lines!(ax, sol.t[j:end], sol[2, j:end])
lines!(ax, sol.t[j:end], sol[3, j:end])
fig


opts_br = BK.ContinuationPar(
    p_min = 0.0, p_max = 7.,
    ds = 0.002, dsmax = 0.01, dsmin = 1e-4, n_inversion = 6,
    detect_bifurcation = 0, max_bisection_steps = 25, nev = 4,
    max_steps = 1000, tol_stability = 1e-3,
)

periodic_orbit_algo = BK.PeriodicOrbitOCollProblem(50, 4; jacobian = BK.DenseAnalyticalInplace(), meshadapt = true, K = 50.)
probpo, cish = BK.generate_ci_problem(
    periodic_orbit_algo, bf_prob, sol, 5.0
)

argspo = (record_from_solution = (x, p;k...) -> begin
		xtt = BK.get_periodic_orbit(p.prob, x, p.p)
		return (max = maximum(xtt[1,:]),
				min = minimum(xtt[1,:]),
				period = BK.getperiod(p.prob, x, p.p),
                p = p.p,)
	end,
)

function plotPO(ax, x, p; ax1=nothing,k...)
	xtt = BK.get_periodic_orbit(p.prob, x, p.p)
	lines!(ax, xtt.t, xtt[1,:]; k...)
	lines!(ax, xtt.t, xtt[2,:]; k...)
	lines!(ax, xtt.t, xtt[3,:]; k...)
end

predictor = BK.PALC(tangent = BK.Bordered())

@time branch = BK.continuation(probpo, cish, predictor, opts_br;
    verbosity = 0, plot = true,
    linear_algo = BK.COPBLS(),
    argspo...,
    bothside = true,
    plot_solution = plotPO,
)

BK.plot(branch)
```

Datseris:
you are right, the period is 5. I don't know why I thought it wraps around 4 times before closing exactly (period 4 orbit). Do you mind making a PR that improves the file with your changes? You only need to change the options that you changed in your script here.

But also I want the continuation to stop at p=4.7, there is no need for this example to go all the way to the homoclinic. We can mention it as a comment "if the continuation is continued with pmin = 0, a homoclinic orbit will be found", or something like that.

rveltz:
OK will do.

Here is the full plot btw (2 homoclinic solutions)

Screenshot 2025-03-11 at 11 48 37 PM

Datseris:
Is this two different initial conditions? I don't understand where the curves start from. In any case, I will see when you PR the code.

PR #173, opened 2025-04-04
Fix link to examples, use ref->id.

guitorri:
Without this the generated docs will throw a 404 error when clicking the 'examples' link, say from the index page.


Datseris (approved):
Perfect thanks. Please also go to the Project.toml file and increment the patch version (3rd number) so I can release a new release after merging your pr!

Datseris (approved):

PR #174, opened 2025-04-11
critical fix in convergence time of recurrences mapper

Datseris:
We **desperately** need rigorous tests for the convergence time estimation of `AttractorsViaRecurrences`. The fact that I've "fixed" this function 3 times, yet was still giving totally wrong results is a testament to that. I've been sloppy to not write tests and someone (me or someone else) needs to contribute rigorous tests for this function and this mapper.

cc @andreasmorr this function will now give massively different results, and much more correct ones.

### DIFF
--- a/src/mapping/recurrences/attractor_mapping_recurrences.jl
+++ b/src/mapping/recurrences/attractor_mapping_recurrences.jl
@@ -185,7 +185,7 @@ function convergence_time(m::AttractorsViaRecurrences)
     else
         x = get(kw, :consecutive_attractor_steps, 2)
     end
-    x = max(i, x) # it cannot be more than i!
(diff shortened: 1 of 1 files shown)


awage:
I will try to design some tests for this as soon as I can.

Datseris:
I have thought of something as well: a discrete dummy map that has dynamics 
```
if x > 0.5
xn = x - 0.5
else
xn = 0
end
```
this way it approaches 0 with fixed steps of 0.5; we make a range `0:1:100` and start with initial condition `x = n`. We know it takes `n` steps to converge, so take into account the required recurrences attractor locate steps (that the function `convergence_time` supposedly already takes into account) and then we have a guaranteed on what the convergence time should be.

PR #175, opened 2025-04-16
new example for an interesting matching procedure

Datseris:
I had to use this for my paper and I thought it was a really nice and elegant way to achieve the matching I wanted. So it is likely useful as an example here?

Someone review and merge please!

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -812,8 +812,8 @@ fig
 ## Matching limit cycles and fixed points in a system with heterogeneous state space
 
 This example discusses the situation of a dynamical system that during a global continuation
-transitions from a fixed point to a limit cycle and then to another fixed point that is on
-the other end of the limit cycle. In the context of this scenario, we want to NOT
+it from a fixed point `A` to a limit cycle `B` and then to another fixed point `C` that is far away (in statespace) from `A`.
+In the context of this scenario, we do NOT want to
 match the fixed points with the limit cycle during the continuation.
 Furthermore, this particular dynamical system has a heterogeneous state space:
 the different dynamic variables have wildly different units, and there is no sensible transformation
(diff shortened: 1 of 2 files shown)


awage:
Tomorrow I'll have a look.

docs/src/examples.md
```diff
@@ -750,61 +750,106 @@ edge states also in high-dimensional and chaotic systems where a simple computat
...
 ax = Axis(fig[1,1], xlabel="x", ylabel="y")
 scatter!(ax, v[:,1], v[:,3]; markersize = 3)
 fig
 ```
+
+## Matching limit cycles and fixed points in a system with heterogeneous state space
+
+This example discusses the situation of a dynamical system that during a global continuation
+transitions from a fixed point to a limit cycle and then to another fixed point that is on
+the other end of the limit cycle. In the context of this scenario, we want to NOT
```
awage, line 816: When you say "on the other end", you refer to the parameter scale? it is not clear.

docs/src/examples.md
```diff
@@ -750,61 +750,106 @@ edge states also in high-dimensional and chaotic systems where a simple computat
...
+match the fixed points with the limit cycle during the continuation.
```
awage, line 817: "we do *NOT* want to match"

docs/src/examples.md
```diff
@@ -750,61 +750,106 @@ edge states also in high-dimensional and chaotic systems where a simple computat
...
+the different dynamic variables have wildly different units, and there is no sensible transformation
+that would bring all variables to the same units.
+
+We will showcase how one can achieve match attractors in this system simply by defining a
+special distance function that is given to [`MatchBySSSetDistance`](@ref). This is:
+
+```@example MAIN
+function centroid_and_length(A, B)
+    # first check we are comparing a fixed point and limit cycle. We do this by
+    # checking if there are different lengths and one of the two is 1 (fixed point)
```
awage, line 828: "We do this by checking if the lengths of sets A and B are different and if one of the sets has length 1 (fixed point)"

docs/src/examples.md
```diff
@@ -750,61 +750,106 @@ edge states also in high-dimensional and chaotic systems where a simple computat
...
+    if length(A) != length(B) && any(isequal(1), length.((A, B)))
+        return Inf
+    end
+    # otherwise we use a weighted centroid distance
```
awage, line 832: "If the check fails, both sets are similar in nature (both limit cycle or fixed points). We return the weighted centroid distance."

docs/src/examples.md
```diff
@@ -750,61 +750,106 @@ edge states also in high-dimensional and chaotic systems where a simple computat
...
+    weights = (300.0, 1.0, 1200.0, 300.0, 10.0)
```
awage, line 833: These weights are important?  300 is repeated twice. I suppose it is related to the heterogeneity issue. It should mentioned.

awage: I just saw it is described bellow. A small comment here doesn't hurt.

awage:
I'm done. Just some clarifications on the code.

docs/src/examples.md
KalelR: a few more sentences to explain this would help, or a schematic maybe; i understood "on the other end" to mean location in state space

KalelR:
> I'm done. Just some clarifications on the code.

Same! Agree with Alex, and cool example!

KalelR (approved):
Cool example! Agree with Alex's suggestions!

Datseris:
thanks!

PR #176, opened 2025-04-24
One more stability measure: mean critical shock

Datseris:
I was wondering, what sort of measure we could use for the "average" or "expected" basin size. Something that more accurately corresponds to basin diameter. It isn't the minimal shock, nor the noncritical. I think this measure is a better quantity for it


andreasmorr:
I like the concept of the mean critical shock but I think this would be more difficult to implement than this no? Because what you proposed in the code is the critical shock averaged over the alternative basins of attraction. And you would rather like to have the mean distance of the attractor to its own basin boundary no?

andreasmorr:
So that would be the distance of the attractor to its own basin points averaged over all nonzero measure basin points? Yes I think that would be a nice spatial metric of the basin structure!

andreasmorr:
I tried looking into the errors of the build but couldn't find the culprit. I think the new version of the constructor of StabilityMeasuresAccumulator is not working correctly.

andreasmorr (approved):

Issue #177, opened 2025-05-21
Documentation: Clarify AttractorsViaRecurrences `(sparse=false)` state modification by `basins_of_attraction` and impact on subsequent `mapper(ic)` calls


k1m9l:
When using `AttractorsViaRecurrences` with `sparse = false`, there's a potentially confusing interaction between `basins_of_attraction(mapper; ...)` (the specialized method for `AttractorsViaRecurrences`) and subsequent calls to `mapper(ic)` (which are used by `basins_fractions(mapper, ...)`). This can lead to unexpected label categorization in the output of basins_fractions.

Specifically, after `basins_of_attraction` correctly identifies and stores attractors (e.g., with labels 1, 2, 3), a subsequent call to `basins_fractions` using the same populated mapper instance may return a dictionary of fractions with different, re-categorized labels (e.g., 0, 1) instead of the original attractor labels (1, 2, 3). However, `extract_attractors(mapper)` will still correctly report the initially found attractors with labels 1, 2, 3.

This discrepancy can be confusing as it appears the mapper's interpretation of its own attractors changes based on its history.

**Example Code**

```
using Attractors, StaticArrays, Printf

# Define Newton map
function newton_map(z, p, n)
    z1 = z[1] + im*z[2]
    # Safer Newton step: check for zero derivative
    df_val = p[1]*z1^(p[1]-1)
    if abs(df_val) < 1e-12 # Avoid division by zero or very small numbers
        return SVector(real(z1), imag(z1))
    end
    dz1 = (z1^p[1] - 1) / df_val
    z1_new = z1 - dz1
    return SVector(real(z1_new), imag(z1_new))
end

ds = DiscreteDynamicalSystem(newton_map, [0.1, 0.2], [3.0]) # p=3 (3 roots)
xg = yg = range(-1.5, 1.5; length = 100) # Coarser grid for MWE
grid = (xg, yg)

# 1. Initialize mapper
mapper_newton = AttractorsViaRecurrences(ds, grid;
    sparse = false, # Crucial for this behavior
    consecutive_lost_steps = 1000
)

# 2. Populate mapper with basins_of_attraction
# This version of basins_of_attraction modifies mapper_newton.bsn_nfo.basins
println("Running basins_of_attraction...")
basins_grid_output, attractors_from_boa = basins_of_attraction(mapper_newton; show_progress = false)
println("Attractors from basins_of_attraction (should be e.g., 1, 2, 3):")
display(attractors_from_boa)

[...]

PR #178, opened 2025-05-21
fix basins_of_attracion action on internal variables of the mapper

awage:
`basins_of_attraction` modifies the internal of the mapper. This should fix the problem repported in #177.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "Attractors"
 uuid = "f3fd9213-ca85-4dba-9dfd-7fc91308fec7"
 authors = ["George Datseris <datseris.george@gmail.com>", "Kalel Rossi", "Alexandre Wagemakers"]
 repo = "https://github.com/JuliaDynamics/Attractors.jl.git"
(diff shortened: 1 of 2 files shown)


Datseris:
This is not correct unfortunately. The whole point of this method is that it is the _original_ method we published back in 21. In this method the basin cells are kept track of and recurrences in the basins themselves may lead to an early return. With the change you do now, this is no longer possible. 

See: https://github.com/JuliaDynamics/Attractors.jl/blob/4697e7da53daa66a9db9923b4a151cf087eca22b/src/mapping/recurrences/attractor_mapping_recurrences.jl#L230-L232 


The fact that tests of this PR pass means that we are not actually testing this algorithm that counts consecutive basin cell hits. Otherwise this PR would cause that test to fail. (But it doesn't exist).

awage:
Right! There is an easy fix, the problem is at the end when the label are changed.

Datseris (approved):
pliease also bump patch number in project,.toml

awage:
No problem!

PR #179, opened 2025-05-22
Test for the bug patch in #177

awage:
The bug described in #177 and patched in #178 had no test. 

This test sets up a simple map, runs `basins_of_attraction` with the `sparse = false`  option and the check whether basins_of_attraction has not compromised the mapper by changing the internal labels. 

I added a fix for the documentation compilation error.


Datseris (approved):

Datseris:
thanks for the doc fix!

PR #184, opened 2025-08-26
Update of StabilityMeasuresAccumulator

andreasmorr:
This refactoring makes the computation of stability measures in the StabilityMeasuresAccumulator framework more resource efficient. It also fixes several bugs, such as problems with empty basins of attraction.

### DIFF
--- a/src/mapping/stability_measures_accumulator.jl
+++ b/src/mapping/stability_measures_accumulator.jl
@@ -206,6 +206,9 @@ function finalize_accumulator(accumulator::StabilityMeasuresAccumulator)
     bs = accumulator.bs
     cts = accumulator.cts
     N = length(u0s)
+    N == 0 && error("No initial conditions have been processed. Cannot finalize accumulator.")
+
+
 
     ws = [pdf(accumulator.weighting_distribution, u0[1]) for u0 in u0s]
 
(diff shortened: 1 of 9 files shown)


src/mapping/stability_measures_accumulator.jl
```diff
@@ -172,15 +160,9 @@ function reset_mapper!(a::StabilityMeasuresAccumulator)
...
-    a.basin_points = Dict{Int, StateSpaceSet{Dims,Datatype}}()
-    a.finite_time_basin_points = Dict{Int, StateSpaceSet{Dims,Datatype}}()
-    a.nonzero_measure_basin_points = Dict{Int, StateSpaceSet{Dims,Datatype}}()
-    a.mean_convergence_time = Dict{Int, Float64}()
-    a.maximal_convergence_time = Dict{Int, Float64}()
-    a.mean_convergence_pace = Dict{Int, Float64}()
-    a.maximal_convergence_pace = Dict{Int, Float64}()
-    a.convergence_times = Dict{Int, Vector{Float64}}()
-    a.convergence_paces = Dict{Int, Vector{Float64}}()
+    a.u0s = Vector{StateSpaceSet{Dims,Datatype}}()
```
Datseris, line 163: Why is this a vector of state space sets? It should just be a vector of points?

Datseris: Just a vector? what `u0` is already.  you can transform to a statespaceset at the end, no need to store it

src/mapping/stability_measures_accumulator.jl
```diff
@@ -127,7 +127,7 @@ refers to the distance established by the `metric` keyword.
 """
 mutable struct StabilityMeasuresAccumulator{AM<:AttractorMapper, Dims, Datatype} <: AttractorMapper
     mapper::AM
-    u0s::Vector{StateSpaceSet{Dims, Datatype}}
+    u0s::Vector{SVector{Dims, Datatype}}
```
Datseris, line 130: If this is `SVector` it doesn't work for high dimensional systems. You need to parameterize this as `Vector{V}` and make `{V <: AbstractVector}` a type parameter of the accumulator

src/mapping/stability_measures_accumulator.jl
```diff
@@ -145,7 +145,7 @@ function StabilityMeasuresAccumulator(mapper::AttractorMapper;
     AM = typeof(mapper)
     StabilityMeasuresAccumulator{AM, Dims, Datatype}(
         mapper,
-        Vector{StateSpaceSet{Dims, Datatype}}(),
+        Vector{SVector{Dims, Datatype}}(),
```
Datseris, line 148: initialize this as `Vector{typeof(current_state(ds))}()`.

src/mapping/stability_measures_accumulator.jl
andreasmorr: Nevermind I got it

Issue #187, opened 2025-10-11
Formalize basins with dedicated types


Datseris:
**Describe the feature you'd like to have**

At the moment we handle basins of attraction as a standard Julia Array. The grid corresponding to the basins is a separate variable, and the attractors are as well. This has two major problems:

1. It is tedius as one has to carry around the grid and attractors and basins.
2. It does not play well with an alternative formulation of basins that one gets from `basins_fractions`: the sampled basins. There the basins are not an Array but rather a vector of points. 

I think formalizing this concept is a benefit, and it also allows us to extend it naturally and simply to future research.

**If possible, sketch out an implementation strategy**

- Define abstarct type `BasinsOfAttraction{ID}`. Always "contains" attractors. Type parameterization `ID` is what is the thing stored as an attractor label. For standard basins this is `Int`.
- Define subtype `ArrayBasinsOfAttraction`. Contains a `grid` (instance of `Grid` subtype), and the basin array.
- Define subtype `SampledBasinsOfAttraction`. Contains two vectors:  `sampled_points` and `points_ids`. Initial conditions and their labels.

We can then directly pass the instance of such types into functions like final state sensivitty, or the `heatmap_basins_attractors`, or in general all plotting functions.

We can then extend `ProbabilisticBasinsOfAttraction` where `ID` is not `Int` but rather `Dict{Int, Float}`, the probability to go to each attractor.

PR #189, opened 2025-11-05
Dedicated Basin of Attraction types

JamesBluck:
#### Creation of dedicated basins types: `BasinsOfAttraction` and the sub-types `ArrayBasinsOfAttraction`, `SampledBasinsOfAttraction` as detailed in #187.

- `BasinsOfAttraction{ID}` an abstract type with parameter `ID` which is the type of the attractor label or more generally the type of "each element" of the basin, typically an `Int`. Extending this interface involves creating a sub-type and implementing 3 functions `extract_basins`, `extract_attractors`, and `extract_domain` which access the 3 key fields.
- `ArrayBasinsOfAttraction{ID}` sub-type of `BasinsOfAttraction{ID}` containing:
- - `basins`: An array with element type `ID`, the result of `extract_basins`
- - `attractors`: A dictionary mapping attractor labels to state space sets, the result of `extract_attractors`
- - `grid`: An instance of a sub-type of `Grid` for which the basin was calculated for, the result of `extract_domain`
- SampledBasinsOfAttraction{ID} sub-type of `BasinsOfAttraction{ID}` containing:
- - `points_ids`: A vector with element type `ID`, the result of `extract_basins`
- - `attractors`: A dictionary mapping attractor labels to state space sets, the result of `extract_attractors`
- - `sampled_points`: A state space set of length equal to that of `points_ids` for which the vector basin / labels was calculated for, the result of `extract_domain`

Doc strings have been added for these.

#### An aside on the load order
The `ArrayBasinsOfAttraction` type requires `Grid` to be defined which is included in `mapping/recurrences` this is initially fine as the load order is mapping -> basins, however several functions and types in the mapping folder e.g. BasinsInfo, and basins_of_attraction will require ArrayBasinsOfAttraction once they have been modified. I think the cleanest solution to this with minimal changes to the load order or general file structure is to have the new basin types in a new file, `basins_types.jl` potentially in `/basins` or perhaps `/mapping`, with all the constructors which might rely on several parts of `/mapping` to be located in `basins_utilities.jl` so that it is loaded after mapping. Then if we move the definition of abstract type `Grid` from `attractor_mapping_recurrences.jl` we can then include `basins_types.jl` before the recurrences stuff is loaded without conflict. This is definitely not the only way, if anyone had a better idea I'd like to hear it!

#### Constructors
Most of the constructors for the new types are found in `basins_utilities.jl`,  in particular all the new ones are. The ones that are the result of a modification to existing functions are:

- `basins_of_attraction(mapper::AttractorsViaRecurrences; show_progress = true)` in `attractor_mapping_recurrences.jl` returns an instance of `ArrayBasinsOfAttraction`
- `basins_of_attraction(mapper::AttractorMapper, grid::Tuple; kwargs...)` in `basins_utilities.jl` returns an instance of `ArrayBasinsOfAttraction`

Defining these `extract_...` functions allows us to make `BasinsOfAttraction`, in general, iterable which is essential for ensuring the backwards compatibility for functions that used to return `basins, attractors` like in the above `basins_of_attraction` functions.

Note that the function `convergence_and_basins_of_attraction` in `basins_utilities.jl` is not able to be converted in the same iterable manner as it returns 3 items: basins, attractors, iterations. So wrapping the first two in one of the new types would form a 2-tuple of the form ((basins, attractors), iterations). I have noted this in the doc string. 

Doc strings have been added / modified accordingly.

#### Wrapper functions
Wrapper functions have been created allowing for compatibility with these new types on all basins-related functions, whilst preserving backwards-compatibility, here is a list:
- In `ext/plotting.jl`: `heatmap_basins_attractors`, `shaded_basins_heatmap`, and their in-place versions
- In `basin/`: `basins_fractions`,`basin_entropy`, `basin_fractal_test`, `basin_fractal_dimension`, `uncertainty_exponent`, `test_wad_merge`
- In `matching/basin_overlap.jl`: `matching_map` and the in-place version
- In `tipping/tipping_probabilities.jl`: `tipping_probabilities`

All the documentation strings have been modified to include these wrapper functions.

#### Recurrences
The `BasinsInfo` structure has been modified to contain an instance of the `ArrayBasinsOfAttraction` instead of carrying `basins`, `attractors`, `grid_nfo` separately. Accordingly all references to those fields in `attractor_mapping_recurrences.jl` and `finite_state_machine.jl` have been updated, in addition to a minor reference in `docs/src/recurrences_animation.md`.

#### Interpolating basin functions
Also detailed in a comment in #187, the function `map_to_basin` has been defined in `basins_utilities.jl` for the 2 basin of attraction sub-types.

#### Still To-Do
Add tests

### DIFF
--- a/src/basins/basins_types.jl
+++ b/src/basins/basins_types.jl
@@ -32,13 +32,6 @@ was done to ensure backwards compatibility for functions whose original return f
 
 The [`map_to_basin`](@ref) function provides simple interpolation of a point in state space 
 to determine which basin of attraction it is likely to belong to.
-
-## For developers
-
-`BasinsOfAttraction` defines an extendable interface. A new type needs to subtype
-`BasinsOfAttraction` and implement `extract_basins`, `extract_attractors`,
-and `extract_domain`. To use the [`map_to_basin`](@ref) function, the internal
-function `map_to_domain` must be implemented
 """
 abstract type BasinsOfAttraction{ID} end
 
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -74,6 +74,13 @@ ArrayBasinsOfAttraction
 SampledBasinsOfAttraction
 ```
 
+The `map_to_basin` function provides simple interpolation of a point in state space to determine which basin of attraction it
+is likely to belong to:
+
+```@docs
+map_to_basin
+```
+
 Calculating basins of attraction, or their state space fractions, can be done with the functions:
 - [`basins_fractions`](@ref)
 - [`basins_of_attraction`](@ref)
--- a/src/mapping/attractor_mapping_proximity.jl
+++ b/src/mapping/attractor_mapping_proximity.jl
@@ -115,7 +115,7 @@ function _deduce_ε_from_attractors(attractors, search_trees, verbose = false)
                 k == m && continue
                 for p in A # iterate over all points of attractor
                     Neighborhood.NearestNeighbors.knn_point!(
-                        tree, p, false, dist, idx, Neighborhood.NearestNeighbors.always_false
+                        tree, p, false, dist, idx, Returns(false)
                     )
                     dist[1] < minε && (minε = dist[1])
                 end
--- a/src/mapping/grouping/nearest_feature_config.jl
+++ b/src/mapping/grouping/nearest_feature_config.jl
@@ -49,7 +49,7 @@ end
 # has the same performance!
 @inbounds function feature_to_group(feature, config::GroupViaNearestFeature)
     (; tree, max_distance, dummy_idxs, dummy_dist) = config
-    skip = Neighborhood.NearestNeighbors.always_false
+    skip = Returns(false)
     sort_result = false
     Neighborhood.NearestNeighbors.knn_point!(
         tree, feature, sort_result, dummy_dist, dummy_idxs, skip


Datseris:
@JamesBluck I'll review this, in the meantime can you have a look at why tests fail?

JamesBluck:
Tests should pass now @Datseris

src/basins/basins_types.jl
```diff
@@ -0,0 +1,155 @@
...
+    SampledBasinsOfAttraction,
+    extract_basins,
+    extract_basins,
+    extract_domain
+    
+#########################################################################################
+# Basins of Attraction structure definition
+#########################################################################################
+"""
+    BasinsOfAttraction{ID}
```
Datseris: Throughout all documentation strings, hide the `ID` information. The user doesn't need to know about it at this stage.

src/basins/basins_types.jl
```diff
@@ -0,0 +1,155 @@
...
+    points_ids::Vector{ID}
+    attractors::Dict{AK, S}
+    sampled_points::S
+end
+# The definition of the first constructor can be found in `basins/basins_utilities.jl`.
+
+#########################################################################################
+# Basins of Attraction Convenience functions
+######################################################################################### 
+Base.iterate(BoA::BasinsOfAttraction, state=1) = state == 1 ? (extract_basins(BoA), 2) : state == 2 ? (extract_attractors(BoA), 3) : nothing
```
Datseris: Yes, this is great. this way, the `basins_of_attraction` function can return `BoA` types and not be breaking.

src/basins/basins_types.jl
```diff
@@ -0,0 +1,155 @@
...
+    extract_domain
+    
+#########################################################################################
+# Basins of Attraction structure definition
+#########################################################################################
+"""
+    BasinsOfAttraction{ID}
+
+A subtype of `BasinsOfAttraction` is a convenient structure that stores a representation
+of the `basins` of attraction, their associated `attractors`, and a representation of the `domain`
```
Datseris: Why is the domain here in monotype font? Monotype font refers to a code entity but `domain` doesn't exist in code.

src/basins/basins_utilities.jl
```diff
@@ -155,3 +220,33 @@ function convergence_and_basins_fractions(mapper::AttractorMapper, ics::ValidICS
...
+    n = map_to_domain(BoA, point; kwargs...)
+    return extract_basins(BoA)[n]
+end
+
+# Map point to index of the nearest member of the domain of the basin of attraction
+map_to_domain(BoA::ArrayBasinsOfAttraction, point) = basin_cell_index(point, BoA.grid) # Nearest grid cell
+function map_to_domain(BoA::SampledBasinsOfAttraction, point; tree = KDTree, metric = Euclidean(), ss_kwargs...) 
+    ss = searchstructure(tree, BoA.sampled_points, metric, ss_kwargs...) # Nearest sampled point
+    idx, _ = nn(ss, point, args...; kwargs...)
+    return idx
```
Datseris: This is uber inefficient :D because constructing the KD tree is the most expensive part of the computation. I think, we should construct the KDtree and store it as a field in the _construction_ of the Sampled basins type. Then, remove the option of tree and metric as arguments. They are options for the constructor of the sampled basins instead. We just always initialize a tree even if it is not required.

Later this can be optimized by having the tree field undef, and initializing it on demand if this function is called for the first time, but this can be done later.

src/basins/fractality_of_basins.jl
```diff
@@ -78,11 +89,15 @@ end
...
 """
     basins_fractal_test(basins; ε = 20, Ntotal = 1000) -> test_res, Sbb
+    basin_entropy(BoA, ε = 20, Ntotal = 1000) -> test_res, Sbb
 
 Perform an automated test to decide if the boundary of the basins has fractal structures
 based on the method of Puy et al. [Puy2021](@cite).
 Return `test_res` (`:fractal` or `:smooth`) and the mean basin boundary entropy.
 
+The second function signature is a simple wrapper of the first allowing compatibility with 
+the [`ArrayBasinsOfAttraction`](@ref) type
```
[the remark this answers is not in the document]
Datseris: Repeat this for all functions in this file.

JamesBluck:
I've changed them all to `Returns(false)`, tests should pass now

Datseris:
thanks @JamesBluck , fantastic work!

PR #190, opened 2025-11-15
feat: Add axislegend_kwargs to axislegend calls

oameye:
add the option to move the position of the axislegend


Datseris (approved):
thanks, can you please bump patch version in Project.toml?

PR #191, opened 2025-11-26
use current parameters in jacobian...

Datseris:
@andreasmorr can't believe we've missed this hahahaha, is our preprint affect by this?


andreasmorr:
Oh damn hahah! That's curious though because I am very sure that our results would look very different if indeed the initial parameters were considered for every jacobian computed. That would mean that the only change over parameter values would come from the changing location of the attractor and we wouldnt get the expected divergence of the char. return time for example. Could it be that we were lucky and the ds was initiated with the new parameter value in every step of the continuation?

PR #203, opened 2026-03-28
Add compat for Optim v2

sebapersson:
Basically the same as in https://github.com/JuliaDynamics/ChaosTools.jl/pull/361

Should I add an issue as well?


Datseris:
It's okay I trust you if you say you run test locally and they pass. I run windows so I can fix this on my end some time in the future. thanks for checking this out!

sebapersson:
Thank you, I have also made an associated issue on moving to Optimization

PR #205, opened 2026-04-14
Add new measure: intermingledness

Datseris:
This PR adds a new stability measure called "intermingledness" that I've developed as part of a recent article targetting high dimensional data,

https://arxiv.org/abs/2604.09661

The PR adds the measure as a standalone function, and as a component of the stability measures accumulator. It adds extensive tests and an example. 

The PR also reorganizes the folder structure a bit.


Datseris:
@andreasmorr or @awage I am wondering if either one of you has the time to review this? Let me know if you don't that's okay, I think the PR is in a good state to be mergable as is, but it would of course be better to have a review as well.

PR #206, opened 2026-05-12
Aggregate attractors before computing stability measures

andreasmorr:
This PR adds an optional `aggregation` keyword argument to `finalize_accumulator` that allows merging multiple attractors into a single attractor before computing stability measures.                                                           

`aggregation` is provided as a `Dict` mapping new attractor IDs to vectors of existing IDs (e.g. Dict(1 => [1, 3], 2 => [2])). The new function `_apply_aggregation` merges the specified attractors and their basins before computing all nonlocal stability measures. This is useful when attractors have a common physical interpretation, or when one simply wants to treat a set of attractors as a single entity. Linear stability measures are always NaN for merged attractors since they are not fixed points.

A validation step ensures every attractor ID appears in exactly one group; an error is thrown otherwise. Unused IDs in the aggregation values (i.e., IDs not present among the current attractors) are silently ignored, which allows using a fixed aggregation dict across continuation steps even when not all attractors exist at every parameter value. It is also possible to pass a vector of aggregation dicts or a function of a parameter to the continuation of stability measures, to allow different aggregations during the continuation.


Datseris:
Can we please make this new functionality use the same syntax as https://juliadynamics.github.io/DynamicalSystemsDocs.jl/attractors/stable/api/#Attractors.aggregate_attractor_fractions  and cross reference that docstring? 

That is, aggregation should happen based on the inputs `featurizer, group_config`. This establishes coherence of functionality. In addition, aggregation happens on the basis of the actual attractor properties, which is the thing a user "obserrves" (or cares about). The current suggestion of `Dict(1 => [1, 3], 2 => [2])). ` has no context meaning, as attractor IDs are arbitrary, and also a user cannot know what id each attractor has prior to finalization, as it is the finalization that finally returns the attractors.

Datseris:
> A validation step ensures every attractor ID appears in exactly one group; an error is thrown otherwise. Unused IDs in the aggregation values (i.e., IDs not present among the current attractors) are silently ignored, which allows using a fixed aggregation dict across continuation steps even when not all attractors exist at every parameter value. It is also possible to pass a vector of aggregation dicts or a function of a parameter to the continuation of stability measures, to allow different aggregations during the continuation.

This also won't be necessary when using the existing aggregation infrastructure, as the grouping takes care of all of these things, and also assigns the `-1` special ID to attractors that couldn't be grouped (aggregated).

Datseris:
I'm sorry, I should have told you of this functionality before you put in the work on this PR. I thought you were aware of it, in the call I was suggesting to just re-use the code of the `aggregate_attractor_fractions` function but I wasn't clear enough about it.

andreasmorr:
No worries! Is this what you had in mind?

Datseris:
Hi @andreasmorr , perfect thanks. I did some minor changes. I also had to do a major change: if we are aggregating, we must return the aggregated attractors, as it cannot known otherwise which attractors have been merged with which others... And also the dictionary IDs contained in the internal mapper (like e.g., the recurrences one) are totally irrelevant now; only tyhe aggregator attractors ID matters, as these are the IDs the stability measure dictionaries also have. 

This unfortunately means that the return value of `finalize` must change, so that it returns both things. This also means that in all functionalty that utilizes this function, we must change the syntax and make sure it works. And we must also use the attractors that this function returns instead of using the attractors of the internal mapper, which is what `extract_attractors` returns. 

Are you okay with doing these final changes?

Datseris:
(the codebase is now broken as everywhere we write `measures = finalize(accumulator)` this doesn't work anymorte)

andreasmorr:
Hey George, ah yes of course, that makes sense. I'll make sure it's changed everywhere `finalize` is used.

Datseris (approved):
Great don't forget to add a changelog and increment the package version so that we can tag and release! probably worth mentioning it also in the news in the main `index.md` documentation file.

andreasmorr:
@Datseris could you check whether this last change to the featurizing is sensible? I switched from multi thread to single thread in the featurizing done during one of the tests and an error appeared because `hintsize!` was not available. This fix is a suggestion by an LLM.

Datseris:
yes this is fine!

Datseris (approved):
@andreasmorr do you think it is worth copying here the example you've added in the paper with this aggregated ecosystem model? In the Examples page? It can be run at a smaller parameter resultion like e.g., the 4 parameters values that you show in the basin plots.

Datseris (approved):

PR #217, opened 2026-06-22
Grouped attractor continuations with consistent cross-parameter labels

andreasmorr:
This PR makes it possible to **aggregate/group attractors along a global continuation** and have the resulting groups carry **consistent IDs across the parameter axis**. The same mechanism is used for both the basin-fraction curves and the nonlocal stability measures, so merged attractors and merged measures always share labels.

Previously, grouping was only available per parameter step (e.g. inside `finalize_accumulator`), where each step's groups received arbitrary labels with no relationship across steps. This PR adds the cross-step matching needed to track a group as a single object along the continuation.

## Changes
  - **`MatchByFeatureDistance`** (new, exported): an `IDMatcher` that matches IDs by the Euclidean distance between **feature vectors**. Used with `match_sequentially!`, where the per-step dictionaries map each group ID to its feature centroid (the mean of `featurizer(A)`over the group's member attractors). Consecutive steps are matched by greedy minimum-distance assignment, exactly like `MatchBySSSetDistance` but in feature space.
  - **`aggregate_continuation`** (new, exported): like `aggregate_attractor_fractions`, but also returns the **merged attractor `StateSpaceSet`s per step** alongside the aggregated fractions and the per-group info. All three share consistent group IDs along the curve.
  - **`featurizer` / `group_config` keywords on `global_continuation`** (for `AttractorSeedContinueMatch`): when both are given, attractors are grouped at each step and the groups are matched across parameters by feature centroid, returning aggregated fractions and merged attractors directly.
  - **`stability_measures_along_continuation`**: the grouped output (when `featurizer` / `group_config` are supplied) now has consistent group IDs across parameter steps instead of arbitrary per-step labels.


Datseris:
Isn't this `MatchByFeatureDistance` the same thing as `MatchBySSSetDistance`, but internally in the distance function you first cast each Set to its feature vector and then return the Euclidean distance of these feature vectors........?

Datseris:
It is also confusing to me that `featurizer` is not an input to this `MatchByFeatureDistance`.

Datseris:
This PR will go into a 2.0 release of Attractors.jl so we don't have to worry about breaking changes. I think a single `aggregate_continuation` function is enough. In there, the first  argument could be either the fractions continuation or the stability quantifeirs continuyation. In the first case, the fractions continuation is transformed to exactly the same type oas the stability quantifiers continuation, and then passed to the second function. This minimizes code duplication.

andreasmorr:
Yes these are things we should discuss. As I said in the email we could do without the new matcher, but that would "misuse" the naming of the state space matcher because we are not matching in state space. If we pass the featurizer to the matcher, we would always do the featurizing twice, once at the aggregation stage and then again at the matching stage. Would be fine by me but I wasn't sure if this could be a problem.

Datseris:
At a high level, so that I undersatnd:

1. At each continuation step you are doing a standard featurize + group approach on all found attractors.
2. Between sequential continuation steps you are matching the group centroids using the match by feature distance.

Yes?

Datseris:
> This PR will go into a 2.0 release of Attractors.jl so we don't have to worry about breaking changes. I think a single `aggregate_continuation` function is enough. In there, the first argument could be either the fractions continuation or the stability quantifeirs continuyation. In the first case, the fractions continuation is transformed to exactly the same type oas the stability quantifiers continuation, and then passed to the second function. This minimizes code duplication.

Following up on this, I think we also should NOT do aggregation DURING the global cont of the accumulator (as it was before with the featurizer keyword). It should be done as  a separate function after the fact. This way we have a single function that does the aggregation and it works both for only-fractions and for stability quantifiers continuation.

Thus, `global_cont` should not have featurizer as input.

andreasmorr:
Yes exactly.

Datseris:
> > This PR will go into a 2.0 release of Attractors.jl so we don't have to worry about breaking changes. I think a single `aggregate_continuation` function is enough. In there, the first argument could be either the fractions continuation or the stability quantifeirs continuyation. In the first case, the fractions continuation is transformed to exactly the same type oas the stability quantifiers continuation, and then passed to the second function. This minimizes code duplication.
> 
> Following up on this, I think we also should NOT do aggregation DURING the global cont of the accumulator (as it was before with the featurizer keyword). It should be done as a separate function after the fact. This way we have a single function that does the aggregation and it works both for only-fractions and for stability quantifiers continuation.
> 
> Thus, `global_cont` should not have featurizer as input.

If you think you cna do these changes before wednesday that would be nice, because this simplifies a lot the total amount of code making the review easier. It's okay if you don't have the time though!

andreasmorr:
Yes I agree, I'll give it a shot!

Datseris:
and don't worry about making it backwards compatible

Datseris:
So, final plan:
 
For now, remove possibility of aggregation during continuation. We put all eggs into this 'aggeregate_lalong_continuation' thing.
We have a new function aggregate_attractors_along_continuation. this only takes the attractors_cont, featurizer, group config, and point distance function and threshold. Does matching as well.
New function returhs three thingns. (a) aggregated attractors (union set), (b) the feature centroids continiation, (c) a dinctionary mapping feature centroid IDs to the attractors that became teh group (like Dict(1 => [1, 2]).
 
Just reuse MatchBySSSet distance.

Datseris:
@andreasmorr I thought the function `aggregate_attractor_fractions` wouldn't exist anymore

Datseris:
The file `aggregate_attractor_fractions.jl` is also left unchanged. Can you please clarify, what is the plan now, do we also keep this old stuff?

andreasmorr:
We can remove it but my last status was that you wanted to keep it. But you are right, if we keep it then we should use the new continuation functionality instead of the old way.

andreasmorr:
It's the only measure that can be aggregated post-hoc so there is a case to make for the existence of a way to aggregate it for free after one expensive continuation.

Datseris:
> It's the only measure that can be aggregated post-hoc so there is a case to make for the existence of a way to aggregate it for free after one expensive continuation.

The argument is that by returning `members_cont` in the new aggregation, you can aggregate fractions with only a couple of lines of code. I'll add this now, I am putting in more doc clarificaitons.

andreasmorr:
Ah yes great! Want me to remove the old functions and file then?

Datseris:
Yes, I'll do that. Remember we are moving towards 2.0 so its okay for braking changes.

Datseris:
@andreasmorr why are you defining the parameter curve, and initial conditions twice in the exmaple? You do it once, and then when the stability measures accumulation happens you do it again, and now this time you do

```
ics = [copy(sampler()) for _ in 1:100]
```
instead. Any reason to not re-use what has been defined?

andreasmorr:
I'm not sure I see what you mean. In the first block a prange is needed and in the second I construct the pcurve from that right? And the initial conditions are passed once via a grid and once via sampled ics directly. Is there a way to do these two things consistently in both blocks?

Datseris:
@andreasmorr okay i'm done. see new example. How are the tests looking? I see you've added and adjusted all tests. Do you mind adding one more test for the new `aggregate_fractions`?

We should also make sure that all tests that referred to the old `aggregate_attractor_fractions` are removed.

Datseris:
> I'm not sure I see what you mean. In the first block a prange is needed and in the second I construct the pcurve from that right? And the initial conditions are passed once via a grid and once via sampled ics directly. Is there a way to do these two things consistently in both blocks?

Pass `pcvurve` in both blocks and pass `sampler` in both blocks as I do now.

andreasmorr:
@Datseris I added a test for the fraction aggregation. When running my script for the population model, I encountered two problems, not sure why they don't come up in the tests:

The version StateSpaceSets = "2.5.5" is apparently not available. Could be my local problem though. 

The 2.0 bump makes Attractors incompatible with DynamicalSystems, which requires Attractors = "1".

Datseris:
> The 2.0 bump makes Attractors incompatible with DynamicalSystems, which requires Attractors = "1".

Yes of course, but why is this a problem?

Datseris:
> The version StateSpaceSets = "2.5.5" is apparently not available. Could be my local problem though.

This has to be your own local problem I am afraid because I have the PR checked out with that version of SSS

Datseris:
All tests pass here, that's great! Shall we merge it? Then I'll go ahead with renaming the `AttractorMapper` to `BasinMap`.

andreasmorr:
> > The 2.0 bump makes Attractors incompatible with DynamicalSystems, which requires Attractors = "1".
> 
> Yes of course, but why is this a problem?

Ah all good then, I thought this automatically makes the newest version unusable if you try to load both packages.

andreasmorr:
> All tests pass here, that's great! Shall we merge it? Then I'll go ahead with renaming the `AttractorMapper` to `BasinMap`.

Yes sounds good! Will you also rename the stability measures to quantifiers or should I do that?

Datseris:
> Will you also rename the stability measures to quantifiers or should I do that?

You do that please! Also has to be done in the docs examples!

Datseris:
Holy ffucking hell I just pushed all renames directly into `main` branch. Fuuuuuuuuuuuuuuuuuuuck. Sorry! Please try to do your changes in a PR and I'll now go ahead and see if anything broke on `main`.

PR #218, opened 2026-06-26
Mass mapper rename into basin map

Datseris:
I fucked up and pushed almost everything in `main` branch already, but i'll use this branch to fix all failing tests etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The `AttractorMapper` constructions have been fully renamed. All following renam
 - `AttractorsViaProximity` -> `BasinMapProximity`
 
 This better represents what these constructs are about, which is also reflected in the updated docstring of `BasinMap`.
+Throughout the docs, `mapper` variable has been renamed to `bmap` as well.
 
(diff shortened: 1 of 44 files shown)


Datseris:
Key question: should we also rename `mapper` into `bmap`? `mapper` is the common variable name given to instances of `BasinMap`. It occurs hundreds of times through the docs, in most examples and most docstrings.

Datseris:
> Key question: should we also rename `mapper` into `bmap`? `mapper` is the common variable name given to instances of `BasinMap`. It occurs hundreds of times through the docs, in most examples and most docstrings.

what about this? any thoughts? @andreasmorr ?

Datseris:
@muhammedfadera

andreasmorr:
> > Key question: should we also rename `mapper` into `bmap`? `mapper` is the common variable name given to instances of `BasinMap`. It occurs hundreds of times through the docs, in most examples and most docstrings.
> 
> what about this? any thoughts? @andreasmorr ?

I think it would be nice to rename it.

PR #219, opened 2026-06-26
Rename stability measures to quantifiers

andreasmorr:
Renames the stability *measures* terminology to *quantifiers* throughout the codebase: the `StabilityMeasuresAccumulator` type, the `stability_measures_along_continuation` function, the corresponding source/test files, and all related docstrings and docs. Non-stability uses of "measure" (e.g. `ComplexityMeasures`, measure-theoretic terms) are left untouched.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -207,12 +207,12 @@ swap_dict_keys!
 next_free_id
 ```
 
-## Stability measures/quantifiers
+## Stability quantifiers
 
 ```@docs
-StabilityMeasuresAccumulator
+StabilityQuantifiersAccumulator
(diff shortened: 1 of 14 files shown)


andreasmorr:
@Datseris I saw your message about the erroneous push to main too late. Now I already did all the renaming of measures starting from your status of renamed mappers. Is it ok if we merge the quantifier change into your renaming first and then merge everything into main?

Datseris:
yes we can merge into the renaming brancH!

We have to put the `stability_quantifiers_along_continuation` name into the deprecations.jl file as well. It's an easy backwards compatible change to do!

PR #223, opened 2026-07-26
Introduce sampler interface for `global_continuation`

Datseris:
The third argument to `global_continuation` now has to be an instance of a new type hierarchy, providing instructions on how to sample initial conditions during global continuation. This allow us to straightforwardly add new types in the future. It also takes care of the "messy" code we had so far, where various ways to specify initial conditions were allowed, but never as part of a common interface. 

I am thinking of establishing the same interface for the `basin_fractions` function. The only thing unclear is how to take care of "return labels or not" problem. At the moment, I think I will just create a new function `basin_fractions_labels` or something, as that code was also always messy...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,6 @@ This release accompanies the release of our paper "Global continuation as a comp
   structured around a new abstract type `InitialConditionSampler`.
   - Concrete types include `RandomICSampler, PrescribedICs, PerParameterICs` and more
     to come in the future.
-  - **DEPRECATED**: Passing a vector of initial conditions or a sampling function to
-    `global_continuation` is now deprecated. Use the aforementioned types.
 
 ## Breaking changes and Deprecations
 
@@ -30,6 +28,8 @@ This release accompanies the release of our paper "Global continuation as a comp
 - **DEPRECATED**: giving a vector of initial conditions, or a sampling function,
   as the last argument to `global_continuation` is deprecated. Use an instance of a subtype
   of `InitialConditionSampler` instead.
+- **DEPRECATED**: Passing a vector of initial conditions or a sampling function to
+  `global_continuation` is now deprecated. Use the `InitialConditionSampler` interface.
 
 ### Renaming of stability measures
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ This release accompanies the release of our paper "Global continuation as a comp
   structured around a new abstract type `InitialConditionSampler`.
   - Concrete types include `RandomICSampler, PrescribedICs, PerParameterICs` and more
     to come in the future.
+  - **DEPRECATED**: Passing a vector of initial conditions or a sampling function to
+    `global_continuation` is now deprecated. Use the aforementioned types.
 
 ## Breaking changes and Deprecations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,10 @@ This release accompanies the release of our paper "Global continuation as a comp
   - Concrete types include `RandomICSampler, PrescribedICs, PerParameterICs` and more
     to come in the future.
 
-## Deprecations
+## Breaking changes and Deprecations
 
+- **BREAKING**: The `basins_fractions` output structure has been reworked. Now `basins_fractions` always returns only one output regardless of inputs: the fractions.
+  A new function `basins_fractions_labels` returns also the labels.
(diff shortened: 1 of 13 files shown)


awage:
I though about doing the continuation with this kind of structure: 

```julia
mutable struct BayesianBasinsOfAttraction{R <: HRectangle} <: BasinsOfAttraction{Int}
    alphas::Vector{Dict{Int, Float64}}   # per-tile Dirichlet pseudo-counts
    last_entropy::Vector{Float64}
    last_llr::Vector{Float64}
    attractors::Dict{Int}
    boxes::Vector{R}                     # the tiling 
    β::Float64
end

mutable struct BayesianBasinTracking{A, B <: BayesianBasinsOfAttraction, G, M, S} <: GlobalContinuationAlgorithm
    mapper::A
    boa::B
    sparse_n::Int
    dense_n::Int
    β::Float64
    λ::Float64
    matcher::M
    history_mean_S::Vector{Float64}
    more_history_stuff.... 
end


function global_continuation(bbt::BayesianBasinTracking, pcurve; show_progress = true)
    reset!(bbt)
    mapper = bbt.mapper
    ds = referenced_dynamical_system(mapper)

    Processing stuff


    return fractions_cont, attractors_cont
end

```

Right now, I don't know how we can use the sampler structure to store the information necessary to the algorithm.

Datseris:
hi @awage , see my last commit: https://github.com/JuliaDynamics/Attractors.jl/pull/223/commits/2eb9d16153f578928b3085f4975397ada2cd1809  

this is how I envision it working without the need of a new Continuation Algorithm Type.  Can you write here a list of what things that need to go into the `update_sampler` function?

In the paper you made the decision to further tesselate the state space into boxes, and sample more densely only specific parts of the state space. This is complicated, but also possible. The boxes can be part of the sampler and the sampler can generate initial conditions per-box. The sampler knows which initial conditions correspond to which box by storing for example an enumerator: initial conditions 1-100 are box 1, then 101-200 are box 2, etc.

awage:
I had a look at the code and I am not sure it will fit at this point. I see problems with this approach: 

1) The sampling is adaptive and might need to resample densely at a given parameter step if the log-bayes factor is negative. It means that you have to go back and do the previous step with more samples (call `basins_fraction` again).

2) Basins fractions are not the direct output of the algorithm. You extract the expectation of the dirichlet priors. But this is a minor point that can be adressed.

3) Tesselation is optional but this is a good way to compute only where the boundaries change. 

To answer you question on what's needed: 

The input are the current labels for the box with the respective counts as a dictionnary and the alphas (pseudo-counts) of the Dirichlet distribution (and the parameters of the method).

Datseris:
> 1. The sampling is adaptive and might need to resample densely at a given parameter step if the log-bayes factor is negative. It means that you have to go back and do the previous step with more samples (call `basins_fraction` again).

shiiiiiiiit I forgot about this. This is the only problem. Points 2 and 3 are addressable for sure.

So we have two options: 

1. Allow samplers to re-trigger sampling at a given parameter.
2. Use instead a GlobalContinuationAlgorithm type like you outlined.

Any chance you can do a video call? I think it is worth trying the point 1 because it saves a lot of new code and it allows us to use the accumulator as well.

Datseris:
Alright @awage , after our call, I've finished now the interface around the Sampler API, and made sure that `global_continuation` works with both ASCM and FGAP. It currently doesn't work with StabilityQuantifiersAccumulator but that's fine, I plan to merge this into the existing code.

My proposal now is to merge this PR in, so that I can proceed in a different PR to merge and unify the code of ASCM and StabilityQuantifiersAccumulator using this new output type we discussed. 

In parallel, in a different PR, you can start adding code from your GitHub repo into the functions `generate_ics, update_sampler!` . The function `resampling_required` is thankfully a trivial 1 line of code!

this is looking great!

awage:
Hi @Datseris !  I am working on the `update_sampler!(sampler, labels)` function and there is a small problem with the labels: 

In order to compute the eta factor you have to make sure that the labels from one parameter slice to the following match correclty. So you have to compute and apply the map at each step with `matching_map`. This is in contradiction with the current implementation: the mapping is done at the end with `match_sequentially!`. 

Asolution is to match the attractor inside `update_sampler!` just for the purpose of the adaptive sampling. The all the book-keeping for obtaining the correct labels can be done at the end with a call to a function of the sampler. 

I will think about a way to solve this without changing the continuation code.

Datseris:
We can change the continuation code to do the matching step-by-step. It makes no difference at all for the continuation, and this way the passed labels are matched.

PR #224, opened 2026-07-27
Calculate basin entropy from sampled basins

Datseris:
Also add it in the stability accumulator. Thus, now basin entropy can be calculated during global continuation nearly for free (cost of KDTree can be expensive!)


Datseris:
cc @awage

PR #225, opened 2026-07-29
Dedicated type for output of `global_continuation`

Datseris:
This is hopefully the last change before we can release v2. This is a breaking change and changes the output of `global_continuation`. The reason was that at the moment the same function had different output depending on if it was used with the accumulator or not. This was a bit messy. In addition, there was no way to add more information from the continuation to the output, that may be desired to do in the future as we add more functionality.

So this PR fixes all of that by enforcing the output of global cont to be a special type that has fields that can be extended with more information in the future. The same type is returned regardless of what continuation is being done, which adds coherence to the whole library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ This release accompanies the release of our paper "Global continuation as a comp
   - When continuing with `StabilityQuantifierAccumulator` the third field `.quantifiers`
     contains that info.
   - For backwards compatibility `iterate` works on `GlobalContinuationOutput` and returns
-    the first two fields. This will be breaking user code that used the accumulator though.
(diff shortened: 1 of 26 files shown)


Datseris:
This review has a large amount of changes that makes it impossible to review efficiently, the thing to focus on is this : https://github.com/JuliaDynamics/Attractors.jl/pull/225/changes#diff-2e0bb46db2c5579c4319b4483dd604ef32592a0e0075271a303cf5a5c1d5b8b5R19

Datseris:
I'll merge this in and proceed in establishing the same sampler API for `basins_fractions`, as I feel it is a bit off that it is now a bit incoherent with `global_continuation`!

PR #227, opened 2026-07-31
Minimal implementation of the bayesian sampler routine.

awage:
For now, there is no historical data kept in the sampler. Only the bare minimum. I will draft a minimal working example of the Henon and work on some test. 

The following step is to include the history of the estimated probabilities, but it is only a by-product of the algorithm.


src/continuation/continuation_ascm_generic.jl
```diff
@@ -126,77 +125,54 @@ function global_continuation(
...
+        prev_attractors, prev_p = matched_attractors, p
         # update progress bar
         showvalues = i < length(pcurve) ? [("pcurve index", i + 1)] : []
         ProgressMeter.next!(progress; showvalues)
     end
-    # we now do the matching pass
     rmaps = match_sequentially!(
-        attractors_cont, ascm.matcher; pcurve, ds = referenced_dynamical_system(bmap)
+        attractors_cont, ascm.matcher; pcurve, ds, retract_keys = false
     )
```
Datseris: It is no longer needed to do `rmaps = match_sequentially!(...)`, right ? Since you match the attractors already. What we should do is push the output of `rmap = matching_map` into a vector, and this already creates `rmaps`. 

Also, please notice, this PR is significantly behind the main branch. This function no longer has 
```
match_sequentially!(fractions_cont, rmaps)
return fractions_cont, attractors_cont
```
as its closing statements. 

I think if you pull `main` into your PR it should update it automatically.

awage: Thanks for the comments. I am trying to adapt the matching behavior of match sequentially without breaking stuff. For now it is working but horribly complex. I have to find a solution. Once I have a good way to do the matching, the PR should be ready for review. The sampler interface is done and working.

Datseris: Yes, I think you need to have a look at how the `main` branch is, and start clean. There is no reason to mess with matching fractions inside the code of `glboal_continuation`, we already have all these utility matching functions. 

We can do a call if this helps.

awage:
Application cases are in: https://github.com/awage/BayesianBasinTracking

Datseris:
@awage if you want i have some free time today and i can do some commits on this PR

awage:
Hi! I won’t be able to work on the code until Monday. If you want to commit
some code it’s good for me. The idea is there but it’s based on the old
branch. The critical part is the matching, once it’s solved the sampler
part will be easy.

Datseris:
I propose that you take care of the sampler part and leave the global continuation function to me. Just finish the sampler part, and make sure you tell me what the inputs to the sampler functions are. This should all be in the sampler files. Then I'll make the modifications needed for global continuation in the new code.

Datseris:
@awage What is "weighted fractions"? And do actually we need it in the global continuation code itself? The global continuation generates fractions directly from labels. Is this output of weighted fractions somehow a "better" version of the fractions that are naturally computed from the labels?

awage:
I realized we that we cannot compute the fractions directly if we resample some areas of the phase space: the sampling is no longer uniform and we cannot compute the average fs = counts/total samples. You have to reweight the samples according to the local density. If we use different sampling methods that concentrate the sampling on some areas you will have to do this reweighting.

Datseris:
> I realized we that we cannot compute the fractions directly if we resample some areas of the phase space: the sampling is no longer uniform and we cannot compute the average fs = counts/total samples. You have to reweight the samples according to the local density. If we use different sampling methods that concentrate the sampling on some areas you will have to do this reweighting.

I am not sure this additional complexity is required. In general the user-provided sampling distribution does not have to be uniform to begin with. And we never take this into account but in principle the same argument you present here should be applied to a non uniform sampling function. But okay, this is a different scenario here as we trigger a dense resampling to some other region of the state space and in principle we have the knowledge of where this weighting should occur. So let's keep it.

In any case we can for simplify the code; this whole weighting process should occur on the sampler side. We can add one more function to Interface of samplers like `compute_fractions(labels, ...)` that returns the `fs` instead of doing it inside the global continuation. The default inmpementation of this function is to just calculate the fractions directly from labels by standard weighting, i.e., call the `basin_fractions(labels::Vector)` function directly. But for the Bayesian sampler this will be more complicated. Then this will simplify the global continuation code a lot! Once I have some extra free time i can add some commits here.

Datseris:
i'll have a look tomorrow at this!!!

Datseris:
hi @awage I've made some progress here. The code is definitely a bit clearer. Some key changes contributed to this, for example the addition of `count` functions that return the direct counts. 

There is something however that is unfortunatley non trivial to fix. At the moment, we decided that we must do the matching step-by-step within the continuation, because the samplers will require matched labels. This leads to the following problem: we must take care of either the `vanished` component or the `next_id` component as well. In your version of the code you took care of the `vanished` component but not the `next_id`. It's either-or, as choosing one disables the other.

This overall leads to really ugly code. We have the very elegant function `match_sequentially!`, that takes care of `next_id` and `vanished`, but it only works correctly if the full sequence of the continuation is given. It simply won't work correctly if you use this function with only a single pair of continuation. 

I am currently thinking how we can make this work while keeping an elegant code base. We can write all the `next_id` and `vanished` business within `global_continuation` with a plethora of `if` statements, but this will lead to a total spaghetti code.

p.s. I am on holiday for a week starting tomorrow!

Datseris:
i have an idea how to make things harmonious now

Datseris:
note to self: maybe make retract keys a field of the mapper like use vanished

Datseris:
Right, @awage I have done lots of updates here. Unfortunately i am going on leave now for a week, but please have a look. I am sure I broke some things, but the code is looking super clean now, so this is a good basis to do fixes and improvements if need be!!!

Datseris:
Also, currently the function `weighted_fractions(s::BayesianUpdateSampler, counts)` you have defined does not actually use `counts` anywhere, is this intentional?

awage:
> Also, currently the function `weighted_fractions(s::BayesianUpdateSampler, counts)` you have defined does not actually use `counts` anywhere, is this intentional?

You can use the counts if necessary or use the internal storage of the sampler. When you come back from the break we can discuss the architecture. There are tradeoffs.

Datseris:
I'm back @awage, what's the situation here? I have some minor improvements I can do but they do not relate with architecture. Did you have any chance to look at the current design and perhaps try it? As I've made changes also to the actions of the function, I am not sure if I kept your original functionality intact.

Datseris:
Okay, to clarify, the final code is now in the file `continuation_ascm_generic.jl`.

awage:
> I'm back @awage, what's the situation here? I have some minor improvements I can do but they do not relate with architecture. Did you have any chance to look at the current design and perhaps try it? As I've made changes also to the actions of the function, I am not sure if I kept your original functionality intact.

Hello, today I will pull the code and try some tests.

Datseris:
note that `additive_dict_merge!` is used in other parts of the repo as well and needs to be replaced there as well

awage:
> additive_dict_merge!

Ok, I revert this change.

Datseris:
> Ok, I revert this change.

I meant that we can simply replace all instances of this `additive_dict_merge!` in the library with the `mergewith!(+` version. I didn't know this base function existed!

awage:
I am exhausted, I think the bayesian stuff is OK. I added a new heuristic that improves fine details but that has not much influence on the basins fractions. There are test errors in the nonlocal stuff. Continuation tests pass. I have to pay attention to the bayesian tests, these are llm generated and I should work on it. 

I will recompute the figures of the paper and then we can write down some text.

Datseris:
Okay, thank you for your work!

> There are test errors in the nonlocal stuff. 

I will fix those tests.

> I have to pay attention to the bayesian tests, these are llm generated and I should work on it.

Yes I think double checking LLM output is important. But no rush!

awage:
Tests are failing but it is due to the stability quantifiers.

awage:
You can take it on from here.

Datseris:
thanks @awage . This week its the term start and I am a bit overwhelemd with all the teaching that starts! Please be patient until next week!

